PR #28, opened 2018-07-23
Implement [File] encoding

MihaelIsaev:
I tried to send POST request with `formData` with one file like this
```swift
struct Payload: Codable {
    var key1: String
    var attachment: File
}
```
and it works perfectly, but when I tried to send multiple files like this
```swift
struct Payload: Codable {
    var key1: String
    var attachment: [File]
}
```
it throws an error
`⚠️ [MultipartError.nestedEncode: Nesting is not supported when encoding multipart data.]`

While debugging this issue I found a way how to encode [File] array, and it works perfectly!
I need this functionality in a various of projects and especially in @twof 's Mailgun lib so I hope that this PR will be merged.

### DIFF
--- a/Sources/Multipart/FormDataEncoder.swift
+++ b/Sources/Multipart/FormDataEncoder.swift
@@ -50,6 +50,12 @@ private final class FormDataEncoderContext {
         }
         parts.append(part)
     }
+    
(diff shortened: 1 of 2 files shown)


penny-coin:
Hey @MihaelIsaev, you just merged a pull request, have a coin! 

You now have 81 coins.

PR #54, opened 2021-03-18
Rewrite multipart parser in Swift

siemensikkema:
Rewrites parser in Swift and removes dependency on C library. 

This also fixes an existing parsing bug where a nested multipart body would see its boundaries stripped of the "\r\n--" prefix. (replaces #56)


0xTim:
We probably don't need API docs if we don't expose the Parser. In the same vain I think we can keep the error package protected for now and expose it later if it makes sense. Easier to go that way than the other way!

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -13,192 +13,314 @@ import CMultipartParser
...
+    private let boundary: [UInt8]
+    private let boundaryLength: Int
+    private var state: State
+    private var buffer: ByteBuffer!
+    private var sliceBuffer: ByteBuffer!
 
     /// Creates a new `MultipartParser`.
+    /// - Parameter boundary: boundary separating parts. Must not be empty nor longer than 70 characters according to rfc1341 but we don't check for the latter.
     public init(boundary: String) {
+        precondition(!boundary.isEmpty)
```
0xTim, line 52: Is this safe? If there's a chance we could hit this then we should throw an error instead of bringing down the app

0xTim (approved):
Looks mostly good to me other than the one query around the precondition

0xTim:
Wrt "research better approach for reading and slicing ByteBuffers that doesn't involve a separate copy just for the slicing" maybe @weissi could offer some pointers?

0xTim:
Final comment I think we can roll any improvements into #55 unless anyone spots any blockers. Since all the tests are passing and we've had a few eyes on it I'd be comfortable with merging and tagging

weissi:
> Wrt "research better approach for reading and slicing ByteBuffers that doesn't involve a separate copy just for the slicing" maybe @weissi could offer some pointers?

@siemensikkema What would you need there? `ByteBuffer` can be sliced without copying the bytes either with or without consuming the bytes.

Some examples (all assuming we have a `buffer: ByteBuffer`), none of them copy any bytes:

- `buffer.readSlice(length: foo)`: will conume `foo` bytes from `buffer` and return the consumed portion as a `ByteBuffer`
- `ByteBuffer(buffer.readableBytesView.prefix { $0 != "$" })`: scan the readable bytes until we find a `$` character and construct a `ByteBuffer` which contains all readable bytes until (excluding) the `$` character
- `ByteBuffer(buffer.readableBytesView.drop { $0 == " " }.prefix { $0 != " " })`: trim the `ByteBuffer from leading/trailing spaces

siemensikkema:
@weissi this parser's state machine operates on a byte by byte basis in order to support arbitrary "cuts" in the incoming data which makes decoding from streaming input possible.
But reading one byte at a time the way I do it consumes them so I resorted to keeping another copy of the buffer around which I use to send slices of data via the callback closure once I've determined the offsets.
I need to look into using the techniques you suggested for parsing the body sections while keeping the rest of the parser as is (I don't need slicing for the headers). I _think_ I can get by with a single buffer that way.
But things get tricky with regards to detecting the boundaries when you're never sure whether the current buffer (streamed chunk) suddenly runs out of bytes and what started as a boundary was in fact a part of the body after all.

weissi:
> But reading one byte at a time the way I do it consumes them so I resorted to keeping another copy of the buffer around which I use to send slices of data via the callback closure once I've determined the offsets.

Yup, seems fine. These "copies" won't actually copy the data, so I think you're fine here. `ByteBuffer` is a value type so you can hold onto as many copies as you want, it'll only make a copy when you actually write into the bytes (which you'll never do). Moving reader/writerIndex doesn't count as a write.

> I need to look into using the techniques you suggested for parsing the body sections while keeping the rest of the parser as is (I don't need slicing for the headers). I _think_ I can get by with a single buffer that way.

I think so too but regardless, I wouldn't worry about copies.

0xTim:
@siemensikkema we good to merge and tag then?

siemensikkema:
Original comment before preparing for release:

> Rewrites parser in Swift and removes dependency on C library as part of [this SSWG proposal](https://forums.swift.org/t/multipartkit/45262) and specifically to address [this comment](https://forums.swift.org/t/multipartkit/45262/4).
> 
> This also fixes an existing parsing bug where a nested multipart body
> would see its boundaries stripped of the "\r\n--" prefix.
> 
> - [x] handle illegal characters while parsing headers
> - [x] simplify body parsing
> - [x] add more tests for the parser specifically to test certain edge cases
> - [ ] research better approach for reading and slicing ByteBuffers that doesn't involve a separate copy just for the slicing
> - [ ] add API docs
> - [ ] consider expanding `MultipartParser.Error` with more precise error cases and making it public?

mxcl:
How is the relative performance? Out of interest.

siemensikkema:
@mxcl I want to do a comparison between the old one, the current one and one with a potential performance improvement (using an approach mentioned above) that I have locally because according to #59 the performance has a lot to be desired. 
I didn't get around to setting up the comparison yet.

PR #56, opened 2021-04-09
Fix: Cannot parse data that contains `\r\n--` correctly.

t-ae:
Fix: Cannot parse data that contains `\r\n--` correctly.

### DIFF
--- a/Sources/CMultipartParser/multipartparser.c
+++ b/Sources/CMultipartParser/multipartparser.c
@@ -297,6 +297,7 @@ reexecute:
                     parser->index++;
                     break;
                 }
+                CALLBACK_DATA(data, "\r\n--", 4);
                 CALLBACK_DATA(data, parser->boundary, parser->index);
                 parser->state = s_data;
                 goto reexecute;
--- a/Tests/MultipartKitTests/MultipartKitTests.swift
+++ b/Tests/MultipartKitTests/MultipartKitTests.swift
@@ -6,6 +6,8 @@ class MultipartTests: XCTestCase {
     test123
     aijdisadi>SDASD<a|
 
+    \r\n--
+
     """
 
     let multinamed = """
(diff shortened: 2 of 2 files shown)


t-ae:
When `MultipartParser` finds `\r\n--`, it starts searching boundary.
https://github.com/vapor/multipart-kit/blob/73706f1883f2ba950d41f18aec7e3a53766d4a6d/Sources/CMultipartParser/multipartparser.c#L277-L280

When the sequence turns out not to be boundary, that is a part of data.
`MultipartParser` appends that part to output:
https://github.com/vapor/multipart-kit/blob/73706f1883f2ba950d41f18aec7e3a53766d4a6d/Sources/CMultipartParser/multipartparser.c#L300

It also has to append initial `\r\n--` but it doesn't.
This PR fixes that bug.


While fixing, I found strange test data.
https://github.com/vapor/multipart-kit/commit/70ed0ddd8d46fe423ef1ff06a69ca08dde47d61f
It seems boundary part is corrupted intentionally to hide this bug.
For reviewer: please verify this part.

0xTim:
@t-ae thanks for doing this. Unfortunately we're currently in the process of rewriting the parser to native Swift (see #54 and #55) - can you see if those fix the issue or if you need to add a fix to those instead

0xTim:
@t-ae thanks for doing this. Unfortunately we're currently in the process of rewriting the parser to native Swift (see #54 and #55) - can you see if those fix the issue or if you need to add a fix to those instead

t-ae:
@0xTim 
I saw them before opening this PR.
It seems main PR #54 is work-in-progress and I thought there should be patch version before that.

t-ae:
#54 does very same fix.
https://github.com/vapor/multipart-kit/pull/54/commits/80fe8c994f22539ff78168048c79547d5e11a070#diff-109d8bb6a63edbbebb23d2259c98391f29e3415c41b0a0ea94651645f9a61040L211
So this bug is fixed in #54.

Anyway, it's unclear how long will it take #54 to be merged, I think this PR should be merged at first.

0xTim:
@siemensikkema when do we think #54 will be ready?

siemensikkema:
@0xTim I'd like some input on the remaining open todos (except API docs which are kind of optional since no public API was added). The conclusion might be that both are fine as is but I'd like someone else to have a critical look at it. I'll remove the draft status to better indicate that it's open to reviews.
Otherwise I feel pretty good about it, especially now that @t-ae confirmed my fix 🙂

siemensikkema:
Fixed in #54

PR #57, opened 2021-04-11
Support nested multipart encoding

siemensikkema:
Add encoding and decoding support for nested objects (#57)

In order to avoid overflows by malicious actors there is a maximum nesting depth associated with `FormDataDecoder`. It defaults to `8` but is configurable in the initializer like so: `FormDataEncoder(nestingDepth: 8)`.

This update also deprecates `MultipartError`.


madsodgaard:
@siemensikkema Perhaps we should add a test for encoding/decoding an array for good measure? That was the issue that I faced before this PR

siemensikkema:
> @siemensikkema Perhaps we should add a test for encoding/decoding an array for good measure? That was the issue that I faced before this PR

Absolutely. The current tests for nested data are not sufficient at all and I will expand on them.

0xTim (approved):
Is there any reason to include the 5.3 manifest and pull in Swift Collections if we're copying over all the source?

Other than that this looks great! Have you run the performance tests to see how these compare with all the changes?

0xTim:
Oh if we keep the different manifests we should definitely ensure CI runs on 5.2 and 5.3+ to make sure they work. (I'm fairly sure they do but worth double checking)

siemensikkema:
@0xTim I think we should keep the 5.3 version. The advantages being:
- staying up to date with the latest changes of swift-collections, at least when using 5.3
- more potential for the compiler to optimize if the project referencing this package also includes swift-collections
- semantics: indicate that the 5.2 workaround is really just a workaround and the 5.3 is the "intended" implementation
- trivial to remove once we bump major version: just delete one manifest and the sources folder

re CI: they use the getcidata step and thus run on 5.2 and 5.3.

Tests/MultipartKitTests/MultipartKitTests.swift
```diff
@@ -375,6 +375,129 @@ class MultipartTests: XCTestCase {
...
+        Content-Disposition: form-data; name="bars[][baz]"\r
+        \r
+        3\r
+        -----\r\n
+        """
+
+        let decoder = FormDataDecoder()
+        let foo = try decoder.decode(Foo.self, from: data, boundary: "-")
+
+        XCTAssertEqual(foo, Foo(bar: .init(baz: 1), bars: [.init(baz: 2), .init(baz: 3)]))
```
weissi, line 437: @siemensikkema just a very quick question from a security point of view: What's the maximum nesting depth supported and what happens if we have say 10,000 or 100,000 nested parts?

If I understand this correctly, this code can be triggered from the outside so someone may be sending malicious messages that are deeply nested

siemensikkema: @weissi that might very well lead to a stack overflow since there is no maximum nesting depth. Let me check.

weissi: Yes, if the parser is recursive descent without checking that's likely the case. In that case, you'd need to guard against that because it'd be a trivially exploitable DOS.

weissi: And thanks for checking!

siemensikkema: @weissi yup, and it only took 5566 😞

siemensikkema: @weissi I've restricted the nesting depth to 8 by default but made it configurable through the initializer. Thanks for catching this! I'll be more careful with recursion next time! 😅

Sources/MultipartKit/FormDataDecoder.swift
```diff
@@ -229,6 +229,11 @@ private extension MultipartFormData {
         else {
             return try T(from: _FormDataDecoder(codingPath: codingPath, data: self))
         }
-        return Convertible.init(multipart: part) as! T
+        guard
+            let converted = Convertible.init(multipart: part) as! T?
```
0xTim, line 233: Should this not be
```suggestion
            let converted = Convertible.init(multipart: part) as? T?
```
To avoid crashing if the decoding fails?

siemensikkema: @0xTim the crash was already avoided but you're right, it can be written that way as well. The reason force unwrap works here is that `Convertible.init(multipart:)` returns a `Convertible?` which we know to be `T?`. The previous implementation was doing `as! T` which only worked if the the initializer did not return `nil`.
I now removed the force unwrap and deduplicated this conversion since more or less the same thing was done twice.

Issue #58, opened 2021-04-14
Explore optimizing parser


siemensikkema:
There might be some opportunities for optimizing the new parser from #54 as discussed https://github.com/vapor/multipart-kit/pull/54#issuecomment-817924731 and https://github.com/vapor/multipart-kit/pull/54#issuecomment-818236481.

PR #59, opened 2021-04-17
Improve parsing performance

edwellbrook:
### Context
I noticed the performance when parsing large bodies was a little slow. For a 25MB request body, this was taking ~80s to parse.

Took a little time to naively implement some improvements bringing the parse time for that same 25MB request down to ~6s (so far). I'm sure there's more that can be done, but in the interest of small incremental improvements, those can be addressed in other PRs.

### Changes
- Found that the slowdown was from `ByteBuffer.readInteger`
- Changed from reading a single `UInt8`, to an array of `UInt8`s (of configurable size, defaulting to 1KB)
- Store the read data into `internalBuffer` and an `internalBufferPtr` to represent the current read index in that array
- Read directly from the `internalBuffer` and re-fill when the index reaches the end

### Additional thoughts
There're some similarities with `internalBuffer` and `ByteBuffer` which suggest this could likely be solved by changing how bytes are read from `buffer` (vs. maintaining a separate `UInt8` array). I'm not sure if it's because each call of `ByteBuffer.readInteger` does some checks (i.e. for every byte vs every 1KB chunk) or if it's something else.

Investigating this is a little outside what I can set aside time for. If anyone looks into this and wants to suggest a more straight-forward way of doing this, I'll happily update my PR.

Thanks!

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:5.2
+// swift-tools-version:5.3
 import PackageDescription
 
 let package = Package(
@@ -17,6 +17,6 @@ let package = Package(
             .product(name: "NIO", package: "swift-nio"),
             .product(name: "NIOHTTP1", package: "swift-nio"),
         ]),
-        .testTarget(name: "MultipartKitTests", dependencies: ["MultipartKit"]),
+        .testTarget(name: "MultipartKitTests", dependencies: ["MultipartKit"], resources: [.copy("request-body.txt")]),
     ]
 )
--- a/Sources/MultipartKit/MultipartParser.swift
+++ b/Sources/MultipartKit/MultipartParser.swift
@@ -82,11 +82,15 @@ public final class MultipartParser {
         self.sliceBuffer = buffer
         defer { self.sliceBuffer = nil }
 
+        self.internalBuffer = []
+        self.internalBufferPtr = 0
+        self.offset = 0
+
         try execute()
     }
 
     private func execute() throws {
-        while buffer.readableBytes > 0 {
+        while self.hasReadableBytes() {
             switch state {
             case let .preamble(boundaryMatchIndex):
                 state = parsePreamble(boundaryMatchIndex: boundaryMatchIndex)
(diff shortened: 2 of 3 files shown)


edwellbrook:
Seems I have some tests to fix :-)

siemensikkema:
@edwellbrook Thanks for raising this! It totally slipped my mind to test for performance when rewriting the parser. Lesson learned. This relates to #58 and I think it's worth considering trying the approach referred to in the linked comments there.

siemensikkema:
@edwellbrook by running your test against the parser from 4.0.1 in _release_ mode I get an average of 0.651 seconds. Not super snappy but at least it doesn't turn my laptop into lap heater for minutes on end.
An optimization using `buffer.readableBytesView.prefix { $0 != boundary[0] }` (read a slice until the possible start of a new boundary) makes it go down to a more acceptable (I think) 0.128 seconds. That technique also allows for reducing the complexity quite nicely so I'll try to get that out sometime soon.
I used `swift test -c release -Xswiftc -enable-testing` to get the numbers above. I'll also try comparing with the pre-4.0.1 version.

cc. @mxcl since you commented about performance in #54

siemensikkema:
@edwellbrook got around to testing the c parser and it gets a 0.117 second average. I think we can try inlining some of the helper functions to get the already small difference (0.117 for the c version versus 0.128 for the upcoming optimized swift version) even lower 😌

edwellbrook:
Hey @siemensikkema. Thanks for the comments and for running those release benchmarks.

I'm spending some time today exploring the `prefix` method. Should have an update to the PR a little later if I'm successful. :)

Cheers

siemensikkema:
@edwellbrook I'm very nearly done with that myself so I'll tag you in the PR so you can have a look 🙂

edwellbrook:
@siemensikkema OK. Will just leave it to you then.

With respect — are you looking for contributions from other people?

Feels like I've wasted my time if you just take over.

siemensikkema:
> @siemensikkema OK. Will just leave it to you then.
> 
> With respect — are you looking for contributions from other people?
> 
> Feels like I've wasted my time if you just take over.

@edwellbrook I'm truly sorry to make you feel like you wasted your time! I didn't mean to barge in on your PR that way. I can assure you that you helped a lot by calling attention to the issue (I had optimizing this on my list but I wasn't aware exactly how bad the issue was, even with corrected numbers!), identifying exactly where the problem was and creating the test case!

I guess I couldn't let go of this one because I wrote this parser recently and felt shocked and a bit stupid for introducing the performance degradation. I was very eager to try out the suggestions I received on the PR that introduced the Swift parser and I did mention above that I was nearly done with an improved version.
You also mentioned lack of time in your PR but reading it again it should've been obvious that you were motivated to take it on.

I hope you're not discouraged from making further contributions because I very much look forward to them. I'll make sure to remember to give you and other contributors some space next time!

edwellbrook:
@siemensikkema Appreciate the apology and understand your POV.

Ultimately I'm glad that the perf improvements are happening, regardless of who does them :)

I should say that I'd love to see a solution that improves perf for debug builds as well (if possible). Just checked out your new branch and debug speeds are still in the 80s range, though release builds are indeed better. During development waiting 80s is a little bothersome. Perhaps I just need to test with smaller files for now.

siemensikkema:
I don't have a much better suggestion than to test with smaller files I'm afraid. I'd prefer to stick with the more canonical approach whenever that provides good performance on release builds. Having said that, this difference between debug and release is _quite_ extreme (>600x !).

PR #62, opened 2021-05-12
Fix: MultipartParser cannot decode Non-ASCII header correctly

t-ae:
Fix the bug that `MultipartParser` cannot decode headers that contains non-ASCII characters correctly.


siemensikkema:
@t-ae thanks for the PR! I think restoring the previous behavior is indeed the right solution here. Interpreting that form-charset value can be considered later if needed.

siemensikkema (approved):

PR #66, opened 2021-06-11
Fix regression that broke falling back to Codable

### LINKED ISSUE #65: Decoding String enum
### Describe the bug

Server show the error on a multipart/form-data request with a String enum  - "Could not convert value at "license" to type License from multipart part. for key license". Earlier versions of multipart-kit work as expected. At least 4.0.0.

```
enum License: String, Codable, CaseIterable {
    case dme1
}
```

### To Reproduce

1. Send a multipart/form-data request from a client with a String enum.
2. Try decoding the request in a Content conform struct on the server.
3. Server shows the error.

### Expected behavior

Correct decoding String enum.

### Environment

* Vapor Framework version: 4.47.0
* OS version: Ubuntu 20.04, macOS 11.4
* multipart-kit: 4.2.0

siemensikkema:
Fixes regression that broke falling back to `Codable` for types not conforming to `MultipartPartConvertible` (#66, fixes #65)

Note: this removes conformance of `UUID` to `MultipartPartConvertible` introduced in 4.2.0 because it is no longer needed.

### DIFF
--- a/Sources/MultipartKit/FormDataDecoder.swift
+++ b/Sources/MultipartKit/FormDataDecoder.swift
@@ -228,7 +228,12 @@ private extension MultipartFormData {
 private extension MultipartPart {
     func decode<T>(_ type: T.Type, at codingPath: [CodingKey]) throws -> T where T: Decodable {
         guard
-            let Convertible = T.self as? MultipartPartConvertible.Type,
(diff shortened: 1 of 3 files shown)


0xTim (approved):
LGTM

Issue #67, opened 2021-06-12
Unable to decode nested `multipart` array.


chocoford:
### Describe the bug

```swift
struct Payload: Content {
    struct NestedPayload: Content {
        let name: String
        let age: Int
    }
    let nestedPayload: [NestedPayload]
}

let encoder = FormDataEncoder()
let encoded = try encoder.encode(Payload(nestedPayload: [.init(name: "abc", age: 10),
                                                         .init(name: "def", age: 20)]),
                                 boundary: "-")
let decoder = FormDataDecoder()
let decoded = try decoder.decode(Payload.self, from: encoded, boundary: "-")

```
got output `[ WARNING ] Value required for key 'nestedPayload.0.age'.`

### Environment

* Vapor Framework version: 4.27.0
* Multipart-Kit Framework version: 4.2.1
* Vapor Toolbox version: 18.3.3
* OS version: Mac OS 11.3

PR #69, opened 2021-06-17
Fix the bug of not decoding nested `Array` of Object

chocoford:
Fix the problem that `Multipart-kit` could not decode nested `Array` of Object mentioned in #67.

### DIFF
--- a/Sources/MultipartKit/MultipartFormData.swift
+++ b/Sources/MultipartKit/MultipartFormData.swift
@@ -41,15 +41,14 @@ extension MultipartFormData {
     private static func namedParts(from data: MultipartFormData, path: String? = nil) -> [MultipartPart] {
         switch data {
         case .array(let array):
-            switch array.first {
-            case .keyed:
-                return array.enumerated().flatMap { i, part in
-                    namedParts(from: part, path: path.map { "\($0)[\(i)]" })
-                }
-            default:
-                return array.flatMap {
-                    namedParts(from: $0, path: path.map { "\($0)[]" })
-                }
+            return array.enumerated().flatMap { i, part in
+                namedParts(from: part,
+                           path: {
+                            if case .keyed = part {
+                                return path.map { "\($0)[\(i)]" }
+                            }
+                            return path.map { "\($0)[]" }
+                           }())
             }
         case .single(var part):
             part.name = path
--- a/Sources/MultipartKit/MultipartFormData.swift
+++ b/Sources/MultipartKit/MultipartFormData.swift
@@ -7,10 +7,8 @@ enum MultipartFormData: Equatable {
 
     init(parts: [MultipartPart], nestingDepth: Int) {
         self = parts.reduce(into: .empty) { result, part in
-            print(part.name.map(path))
             result.insertingPart(part, at: part.name.map(path) ?? [], remainingNestingLevels: nestingDepth)
         }
-        print(self)
     }
 
     static let empty = MultipartFormData.keyed([:])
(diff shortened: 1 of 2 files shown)


chocoford:
The situation is more complicated than I thought before.

chocoford:
this brach of `multipart-kit` can decode `form-data` like below. 
```swift
struct Formdata: Decodable, Equatable {
    struct NestedFormdata: Decodable, Equatable {
        struct AnotherNestedFormdata: Decodable, Equatable {
            let int: Int
            let string: String
            let strings: [String]
        }
        let int: String
        let string: Int
        let strings: [String]
        let anotherNestedFormdata: AnotherNestedFormdata
        let anotherNestedFormdataList: [AnotherNestedFormdata]
    }
    let nestedFormdata: [NestedFormdata]
}
```

Fixed that bug that when it comes to `form-data` contains arrays of objects of more than one properties, `Multipart-kit` then decodes it with same object's properties split into different objects(or `keyed` in code).


But `form-data` still lacks the ability to express complex nested structures. For example, when it comes to structure like below:
```swift
struct Formdata: Content {
    struct NestedFormdata: Content {
        let strings: [String]
    }
    let nestedFormdata: [NestedFormdata]
}
```
it cannot distinguish 
```swift
Formdata(nestedFormdata: [
            .init(strings: ["1", "2"]),
            .init(strings: ["3", "4"])
        ])
```
and 
```swift
Formdata(nestedFormdata: [
            .init(strings: ["1", "2", "3", "4"]),
        ])
```
(let me know if I am wrong)

siemensikkema:
@chocoford I'll need to get into the details at a later point when I have more time but with regards to `form-data` not being able to distinguish between the 2 scenarios you mention it seems we'd need to introduce numbered keys. Do you know off-hand if that conflicts with any part of the spec?

chocoford:
> @chocoford I'll need to get into the details at a later point when I have more time but with regards to `form-data` not being able to distinguish between the 2 scenarios you mention it seems we'd need to introduce numbered keys. Do you know off-hand if that conflicts with any part of the spec?

I found `Specifying array keys is optional in HTML. If you do not specify the keys, the array gets filled in the order the elements appear in the form.` in [PHP Documentation](https://www.php.net/manual/en/faq.html.php#faq.html.arrays)

So far, I haven't found any information about that from the `rfc` specifications.

siemensikkema (commented):
It seems to me like we shouldn't try to guess which element of an array a part belongs to but rather require an index except for maybe the last part of the name. So `nestedFormdata[0][strings][]` for example. And we should change the _encoder_ to match that (but there just output indexes for all).

chocoford:
> It seems to me like we shouldn't try to guess which element of an array a part belongs to but rather require an index except for maybe the last part of the name.

That is a great idea. 
Actually when `form-data` brings any array of objects(_in-array-objects_) which contains just one property, the distinguish problem exists.
If all _in-array-objects_ containing multiple properties, the problem mentioned will no longer exist.

So, we have two options: **allow** or **disallow** `optional` indices for non-last keys.
If `allowed`, we must be able to detect if there is an _in-array-object_ only has one property and give warnings.

And I will try to make code modifications to encoder and decoder in next fews days.

chocoford:
So far, I've allowed optional index in non-last part.

chocoford-zhang:
@siemensikkema It doesn't matter, as long as the problem can be solved.

chocoford:
@siemensikkema . It doesn't matter, as long as the problem can be solved.

PR #70, opened 2021-09-01
Redesign FormDataEncoder and FormDataDecoder

### LINKED ISSUE #68: Decoding optional type cause error
### Describe the bug

This one works:
```swift
let foo = try FormDataDecoder().decode(Foo.self, from: data, boundary: "12345")
```

this one doesn't 
```swift
let foo = try FormDataDecoder().decode(Foo?.self, from: data, boundary: "12345")
```

Error:
```
caught error: "typeMismatch(MultipartKit.OrderedDictionary<Swift.String, MultipartKit.MultipartFormData>, Swift.DecodingError.Context(codingPath: [], debugDescription: "expected single value but encountered dictionary", underlyingError: nil))"
```

### To Reproduce

You can adjust `func testFormDataDecoderW3()` to see this behaviour 

### Expected behavior

Content should be decoded no matter if type is Foo.self or Foo?.self 

### Environment

Tested od `mian` branch

siemensikkema:
Redesign FormDataEncoder and FormDataDecoder to better align with Codable's intended behavior.

- Support decoding top level optionals (fixes #68)
- Adds a `userInfo` property to `FormDataDecoder` and `FormDataEncoder`
- More correct handling of "superDecoder" and "superEncoder"
- When nesting depth is exceeded we first fail when we’re actually trying to grab data from beyond the nesting limit
- Add some missing API docs


0xTim (approved):
This looks good - glad to see lot of implementation changes and no test changes. My one concern is that it's hard to ensure that we haven't broken the API with the move to `FormDataDecoder` stuff being in extensions. I _think_ it's ok but just something to bear in mind

PR #71, opened 2021-09-01
Support Indexed Arrays

### LINKED ISSUE #67: Unable to decode nested `multipart` array.

siemensikkema:
Supports indexed array elements when decoding and encoding form data (fixes #67, replaces #69).

Indicating that a value belongs to an array can now be done by appending `[0]`, `[1]`, etc. to the name, as an alternative to the already supported `[]`. This allows for representing nested data _inside_ elements in an array.

For instance, `let matrix: [[Int]] = [[42,21]]` can now be represented using part names `matrix[0][0]` and `matrix[0][1]`. Without indexed array elements there is no way to distinguish the above from `[[42],[21]]`.

> Note: `FormDataEncoder` now also includes indexes for array types in its output.

Credits go to @chocoford for identifying the issue, the test cases and the initial PR!


0xTim (approved):
LGTM

PR #76, opened 2021-09-30
Add `FormDataDecoder.decode` with `ByteBuffer` parameter

### LINKED ISSUE #75: Unnecessary conversions between `[UInt8]` and `ByteBuffer`
### Describe the bug

When decoding a multipart buffer using `FormDataDecoder` you generally start with a `ByteBuffer`. `FormDataDecoder` forces you to convert this to `[UInt8]` which then it converts back to a `ByteBuffer` internally inside `MultipartParser.execute`. This is causing unnecessary data copying.

### Expected behavior

Add
```swift
FormDataDecoder .decode<D: Decodable>(_ decodable: D.Type, from data: ByteBuffer, boundary: String) throws -> D
```
which uses `ByteBuffer` version of `MultipartParser.execute`.

Set other versions of `FormDataDecoder.decode` to use this new version

adam-fowler:
Add `FormDataDecoder.decode` with `ByteBuffer` parameter and use this function in the other `decode` functions. This avoids unnecessary conversions from ByteBuffer to [UInt8] and back.

Add `FormDataDecoder.decode` with `ByteBuffer` parameter

Resolves #75


0xTim (approved):
LGTM

PR #93, opened 2024-01-18
fix typo in README

Patrick-Kladek:
Adding this library to my vapor project fails with error:

```
product 'MultiPartKit' required by package
'connect-server' target 'App' not found in package 'multipart-kit'.
```

in Package.swift the library name is slightly different and once changed the project compiles


0xTim (approved):
Thanks!

PR #95, opened 2024-03-01
Add encoding function that returns Data. Fixes #94

### LINKED ISSUE #94: Encoding corrupts data
### Describe the bug

When using `FormDataEncoder` to encode data objects, specifically JPEG image data, the data gets corrupted.

### To Reproduce

This can easily be reproduced by feeding JPEG image data into `FormDataEncoder`'s encode function, then feeding it's output into `FormDataDecoder`'s decode function and comparing the input data with the output data.

``` Swift
let folder = #filePath.split(separator: "/").dropLast().joined(separator: "/")
let path = "/" + folder + "/Utilities/image.jpeg"
let originalData = try Data(contentsOf: URL(fileURLWithPath: path))

struct ObjectToEncode: Codable {
    let data: Data
}

let object = ObjectToEncode(data: originalData)
let boundary = UUID().uuidString

let encoded = try FormDataEncoder().encode(object, boundary: boundary)
let decoded = try FormDataDecoder().decode(ObjectToEncode.self, from: encoded, boundary: boundary)

XCTAssertEqual(originalData, decoded.data)
```

The comparison fails indicating that the original data is not equal to the decoded data.

When inspecting the bytes it shows a ~90% increase in size. @Joannis suggested one can expect such increase with base64 encoding while @vzsg was more specific pointing out that base64's "expanded size" is 4:3 not 3:2.

@vzsg also expressed concern that the encode method returns a String instead of Data as binary files may not be valid UTF-8.

### Expected behavior

The decoded data is equal to the original data.

### Environment

multipart-kit version: 4.6.0
swift-tools-version: 5.7
macOS version: 14.2.1

andrew804:
Added `encodeToData` function to `FormDataEncoder` which returns Data instead of String.
Added `serializeToData` function to `MultipartSerializer` which returns Data instead of String.
Added `decode` function to `FormDataDecoder` which accepts Data as input.
Added test for encoding and decoding a jpeg image with the new Data functions.
Added test for encoding and decoding a jpeg image with the original String functions.
Added image.jpeg in /Utilities

This PR fixes https://github.com/vapor/multipart-kit/issues/94


Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
```diff
@@ -37,6 +38,22 @@ public struct FormDataDecoder: Sendable {
...
+    ///
+    /// - Parameters:
+    ///   - decodable: Generic `Decodable` type.
+    ///   - data: String to decode.
+    ///   - boundary: Multipart boundary to used in the decoding.
+    /// - Throws: Any errors decoding the model with `Codable` or parsing the data.
+    /// - Returns: An instance of the decoded type `D`.
+    public func decode<D: Decodable>(_ decodable: D.Type, from data: Data, boundary: String) throws -> D {
+        var buffer = ByteBufferAllocator().buffer(capacity: data.count)
+        buffer.writeBytes(data)
```
gwynne, line 53: `import NIOFoundationCompat` and use `ByteBufferAllocator().buffer(data: data)`.

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -23,6 +24,22 @@ public final class MultipartSerializer: Sendable {
...
+    ///
+    /// - parameters:
+    ///     - parts: One or more `MultipartPart`s to serialize into `Data`.
+    ///     - boundary: Multipart boundary to use for encoding. This must not appear anywhere in the encoded data.
+    /// - throws: Any errors that may occur during serialization.
+    /// - returns: `multipart`-encoded `Data`.
+    public func serializeToData(parts: [MultipartPart], boundary: String) throws -> Data {
+        var buffer = ByteBufferAllocator().buffer(capacity: 0)
+        try self.serialize(parts: parts, boundary: boundary, into: &buffer)
+        return Data(buffer.readableBytesView)
```
gwynne, line 41: `import NIOFoundationCompat` and use `Data(buffer: buffer, byteTransferStrategy: .automatic)`.

And while you're at it, change the `String` version to use `String(buffer: buffer)` 🙂

Tests/MultipartKitTests/MultipartTests.swift
```diff
@@ -141,6 +141,42 @@ class MultipartTests: XCTestCase {
         }
     }
 
+    func testFormDataCodingToFromStringWithJPEG() throws {
+        let folder = #filePath.split(separator: "/").dropLast().joined(separator: "/")
+        let path = "/" + folder + "/Utilities/image.jpeg"
```
gwynne, line 146: Please don't do this - if you're going to go to the trouble of adding it to the manifest as a package resource, use `Bundle.module.url(forResource:)` rather than misusing `#filePath` (yes, despite the fact that I advised someone else to do exactly that just last night 😆).

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
```diff
@@ -26,6 +27,20 @@ public struct FormDataEncoder: Sendable {
...
+    /// Encodes an `Encodable` item to `Data` using the supplied boundary.
+    ///
+    ///     let a = Foo(string: "a", int: 42, double: 3.14, array: [1, 2, 3])
+    ///     let data = try FormDataEncoder().encodeToData(a, boundary: "123")
+    ///
+    /// - parameters:
+    ///     - encodable: Generic `Encodable` item.
+    ///     - boundary: Multipart boundary to use for encoding. This must not appear anywhere in the encoded data.
+    /// - throws: Any errors encoding the model with `Codable` or serializing the data.
+    /// - returns: `multipart/form-data`-encoded `String`.
```
gwynne, line 40:
```suggestion
    /// - Parameters:
    ///     - encodable: An `Encodable` item.
    ///     - boundary: The multipart boundary to use for encoding. This string must not appear in the encoded data.
    /// - Throws: Any errors encoding the model or serializing the data.
    /// - Returns: A `multipart/form-data`-encoded `String`.
```

This better matches the style of documentation comments in use elsewhere (or at least anywhere that I've been writing them 😅).

(And of course, update the comments for the other methods you've added accordingly.)

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
```diff
@@ -26,6 +27,20 @@ public struct FormDataEncoder: Sendable {
...
+    public func encodeToData<E: Encodable>(_ encodable: E, boundary: String) throws -> Data {
```
gwynne, line 41:
```suggestion
    public func encodeToData(_ encodable: some Encodable, boundary: String) throws -> Data {
```
Since the package requires Swift 5.7, we can use the more concise form of the generic type here. (If only the syntax was flexible enough to do the same with the decode side of things 😅). The `<E: Encodable>` and `some Encodable` forms are so synonymous that switching between them is not a source compatibility break. (In fact, can you make the same change in the `String` version of the method above?)

gwynne (changes requested):
A great start! Just a few nits, nothing major.

Sources/MultipartKit/MultipartSerializer.swift
andrew804: I used `ByteBuffer(data: data)`  instead but assume that’s the same.

andrew804:
Thank you for your fast feedback! Hopefully I covered everything - please let me know if not or if you have any further comments.

0xTim (approved):
LGTM - @gwynne anything else to add?

0xTim:
@andrew804 Looks like this is failing in CI

andrew804:
@0xTim the failing test is the one that tests for encoding and decoding a jpeg image with the original string functions which highlights the original bug ie the original bug is still present, using the string encoding and decoding functions still corrupts the data.

Therefore I could either 1) add a warning in the documentation to not use the string functions and instead use the data functions for encoding and decoding images, then just remove the failing test or 2) remove the troublesome string functions entirely and include this PR in the next major release.

Unless you have any other suggestion, what would you prefer?

ptoffy:
Hey @andrew804 thanks for this, however I'm closing it as this is fixed in MultipartKit 5. You can simply use the generic encoding function 
```swift
let encoded = try FormDataEncoder().encode(object, boundary: "hello", to: [UInt8].self)
```

PR #96, opened 2024-05-19
Remove no longer needed `@preconcurrency` attribute

gwynne:
`swift-collections` has since been updated to make the attribute unnecessary.


0xTim (changes requested):
Can we bump the package collection to the version where Sendable annotations were added?

gwynne:
> Can we bump the package collection to the version where Sendable annotations were added?

Done

0xTim (approved):

PR #100, opened 2024-10-07
MultipartKit V5

ptoffy:


Sources/MultipartKit/MultipartParser+AsyncStream.swift
```diff
@@ -0,0 +1,45 @@
+public struct MultipartParseSequence: AsyncSequence {
```
adam-fowler, line 1: You could use a generic parameter on the type to avoid the need for AnyAsyncSequence eg
```swift
public struct MultipartParseSequence<Base: AsyncSequence> where Base.Element == ArraySlice<UInt8> {
    let baseSequence: Base
    public init(boundary: String, baseSequence: Base) {
    }
}
```
If you need a type erased version I'm sure you could add a wrapper `AnyMultipartParseSequence`

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,267 @@
...
-    public func execute(_ bytes: [UInt8]) throws {
-        try execute(ByteBuffer(bytes: bytes))
-    }
-
-    public func execute(_ buffer: ByteBuffer) throws {
-        self.buffer = buffer
-        defer { self.buffer = nil }
-
-        try execute()
+    mutating func append(buffer: ArraySlice<UInt8>) {
```
adam-fowler, line 37: Perhaps append can take a generic collection. There is a good chance the source of the Multipart data is a ByteBuffer.
```swift    
mutating func append<Buffer: Collection<UInt8>>(buffer: Buffer) {
    switch self.state {
    case .initial:
        self.state = .parsing(.boundary, .init(buffer))
        ...
```

Sources/MultipartKit/MultipartParser+AsyncStream.swift
```diff
@@ -0,0 +1,45 @@
...
+    private let parser: MultipartParser
+    private let buffer: AnyAsyncSequence<ArraySlice<UInt8>>
+
+    public init<AS: AsyncSequence & Sendable>(boundary: String, buffer: AS) where AS.Element == ArraySlice<UInt8> {
```
adam-fowler, line 5: If you make `MultipartParser.append` take a generic collection, you can change the requirement on `AS.Element` to be `AS.Element: Collection<UInt8>`

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,267 @@
...
-
-        while let byte = readByte() {
-            switch (crlf, byte) {
-            case (.cr, .cr):
-                crlf = .lf
-            case (.lf, .lf):
-                return nil
-            default:
-                throw Error.syntax
+    private mutating func parseBody(from buffer: ArraySlice<UInt8>) -> ReadResult {
```
adam-fowler, line 88: It looks like parseBody only returns complete bodies. is this correct? This won't be very memory efficient if the body is large.

adam-fowler (commented):
Some initial comments.

Sources/MultipartKit/MultipartParser.swift
ptoffy: Not exactly: it returns a body chunk for each line of content. You can see in the tests how that gets parsed. However I'm not sure how multipart encodes data such as images. If it's just a huge string of binary data on one line it might make problems. We'd have to define a chunk size in that case

adam-fowler: You'll need to support binary files. You don't need to define a chunk size though. Just pass the body chunks as you receive them. 

Is your current body parsing code assuming the content is ASCII?

ptoffy: It shouldn't be ASCII specific besides the boundary check

adam-fowler: You are checking for CRLF in data that could be binary. Newlines are not a concept in binary data. The check for CRLF only makes sense if the source data is text.

Is there a reason you want to break up the body into lines?

ptoffy:
@adam-fowler Do you want to take a look at this again and see if stuff makes more sense now? I added in some binary data (which even contains hex-CRLF 😄) to the tests so it should be able to parse anything now

ptoffy:
@Joannis @simonjbeaumont @czechboy0 pulling you into this so you can take a look if you want

Package.swift
```diff
@@ -1,34 +1,33 @@
-// swift-tools-version:5.7
+// swift-tools-version:6.0
```
Joannis, line 1: Is this necessary?

Package.swift
```diff
@@ -1,34 +1,33 @@
...
     name: "multipart-kit",
     platforms: [
-        .macOS(.v10_15),
-        .iOS(.v13),
-        .tvOS(.v13),
-        .watchOS(.v6),
+        .macOS(.v15),
+        .iOS(.v15),
+        .tvOS(.v15),
+        .watchOS(.v8),
```
Joannis, line 10: Are these bump necessary as well? macOS 15 is also a fair bit higher can iOS 15

Package.swift
```diff
@@ -1,34 +1,33 @@
...
     ],
     products: [
-        .library(name: "MultipartKit", targets: ["MultipartKit"]),
+        .library(name: "MultipartKit", targets: ["MultipartKit"])
     ],
     dependencies: [
-        .package(url: "https://github.com/apple/swift-nio.git", from: "2.65.0"),
+        .package(url: "https://github.com/apple/swift-http-types.git", from: "1.3.0"),
```
Joannis, line 16: Love this

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.SingleValueContainer.swift
```diff
@@ -1,34 +1,59 @@
+import Foundation
```
Joannis, line 1: `FoundationEssentials` might be more fitting?

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
```diff
@@ -23,67 +22,36 @@ public struct FormDataDecoder: Sendable {
...
     ///   - decodable: Generic `Decodable` type.
-    ///   - data: String to decode.
+    ///   - data: `String` to decode.
     ///   - boundary: Multipart boundary to used in the decoding.
     /// - Throws: Any errors decoding the model with `Codable` or parsing the data.
     /// - Returns: An instance of the decoded type `D`.
-    public func decode<D: Decodable>(_ decodable: D.Type, from data: String, boundary: String) throws -> D {
-        try decode(D.self, from: ByteBuffer(string: data), boundary: boundary)
+    public func decode<D: Decodable>(_ decodable: D.Type, from string: String, boundary: String) throws -> D {
+        try decode(D.self, from: Array(string.utf8), boundary: boundary)
```
Joannis, line 36: This makes a copy from `string`, can't we use `withContiguousMemoryIfAvailable`?

Sources/MultipartKit/MultipartParser+parse.swift
```diff
@@ -0,0 +1,57 @@
...
+    /// Synchronously parse the multipart data into an array of ``MultipartPart``.
+    public func parse(_ data: Body) throws -> [MultipartPart<Body>] where Body: RangeReplaceableCollection {
+        var output: [MultipartPart<Body>] = []
+        var parser = MultipartParser(boundary: self.boundary)
+
+        var currentHeaders: HTTPFields?
+        var currentBody = Body()
+
+        // Append data to the parser and process the sections
+        parser.append(buffer: data)
```
Joannis, line 13: Is it necessary to copy data out before parsing?

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
ptoffy: Mhh I don't think that would work any better because we need the Collection there. We can't pass in the raw bytes and if we're copying them to an array we're still making a copy at that point right?

Sources/MultipartKit/MultipartParser+parse.swift
ptoffy: This way we avoid a whole bunch of `needMoreData` returns from the parser

Package.swift
ptoffy: With 6.1 being branched I don't think we should go lower than 5.10, and at that point might as well go 6.0. Plus we're currently using `Mutex` which yes _can_ be replaced but why

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
```diff
@@ -38,12 +36,15 @@ public struct FormDataEncoder: Sendable {
     ///     - boundary: Multipart boundary to use for encoding. This must not appear anywhere in the encoded data.
     ///     - buffer: Buffer to write to.
     /// - throws: Any errors encoding the model with `Codable` or serializing the data.
-    public func encode<E: Encodable>(_ encodable: E, boundary: String, into buffer: inout ByteBuffer) throws {
-        try MultipartSerializer().serialize(parts: parts(from: encodable), boundary: boundary, into: &buffer)
+    public func encode<E: Encodable, Body: MultipartPartBodyElement>(_ encodable: E, boundary: String) throws -> Body
```
adam-fowler, line 39: Can we have the generic parameter as a function parameter as well eg
```swift
public func encode<E: Encodable, Body: MultipartPartBodyElement>(
    _ encodable: E, 
    boundary: String, 
    to: Body.Type = Body.self
) throws -> Body
```

Package.swift
Joannis: I noticed a fair amount of people use `latest - 1` still, that's the main _why_. I wouldn't go for 5.9 at this point anyways, but I think if you're using newer features like mutex - that's fair.

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
Joannis: Since you're doing an `.append` on the parser you're right that you're already making a copy. Except right now you're making _two_ copies of the same data, and each time you're also allocating space for that data.

Sources/MultipartKit/MultipartParser+parse.swift
Joannis: Could it be restructured to be `parseOrAppend` then?

```swift
if parserBuffer.isEmpty {
  let result = parser.parse(data)
  if result == .needMoreData {
    parser.append(data)
  }
} else {
  parser.append(data)
  parser.parse()
}
```

Sources/MultipartKit/FormDataEncoder/Storage.swift
```diff
@@ -1,33 +1,56 @@
...
+    }
+
+    var data: MultipartFormData<Body>? {
         dataContainer?.data
     }
 }
 
-protocol DataContainer {
-    var data: MultipartFormData { get }
+protocol DataContainer<Body>: Sendable {
```
adam-fowler, line 18: Why do a bunch of the internal types of the encoder have to conform to Sendable? These are only accessed from a single thread surely.

Sources/MultipartKit/FormDataEncoder/Storage.swift
```diff
@@ -1,33 +1,56 @@
...
-    let data: MultipartFormData
+    let data: MultipartFormData<Body>
 }
 
-final class KeyedDataContainer: DataContainer {
-    var value: OrderedDictionary<String, Storage> = [:]
-    var data: MultipartFormData {
+final class KeyedDataContainer<Body: MultipartPartBodyElement>: DataContainer {
+    private let _value: OrderedDictionary<String, Storage<Body>> = [:]
+    private let box: Mutex<OrderedDictionary<String, Storage<Body>>> = .init(.init())
```
adam-fowler, line 32: If this didn't need to be Sendable you wouldn't need to wrap the value access in a Mutex. You also seem to have a private member `_value` which isn't used

Sources/MultipartKit/FormDataDecoder/FormDataDecoder+Decoder.swift
```diff
@@ -1,15 +1,23 @@
 extension FormDataDecoder {
-    struct Decoder {
+    struct Decoder<Body: MultipartPartBodyElement>: Sendable {
```
adam-fowler, line 2: Does this need to be Sendable?

Sources/MultipartKit/FormDataEncoder/FormDataEncoder+Encoder.swift
```diff
@@ -0,0 +1,38 @@
+extension FormDataEncoder {
+    struct Encoder<Body: MultipartPartBodyElement>: Sendable where Body: RangeReplaceableCollection {
```
adam-fowler, line 2: Does this need to be Sendable?

Sources/MultipartKit/MultipartParserAsyncSequence.swift
```diff
@@ -0,0 +1,71 @@
...
+                    }
+                    parser.append(buffer: next)
+                case .error(let error):
+                    throw error
+                case .finished:
+                    return nil
+                }
+            }
+        }
+    }
```
adam-fowler, line 70: As I understand it the user can receive a body as multiple MultipartSections if the underlying AsyncSequence has broken that body up. This is great as we don't want to pay the memory for large bodies if we can. But there are situations where we want to ensure we have a complete body section eg a block of data we want to run a JSON decode on.  Is it possible to add a helper function to the Iterator to do this? eg `Iterator.collectBody(upTo: memoryLimit)` which returns a header, complete body section

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
ptoffy: I'll reply to the rest when I'm back home but writing to a `ByteBufferView` works

adam-fowler: Yeah that works and you can create a ByteBuffer from that. Make the `to:` parameter even more important as it makes people consider what their output buffer should be.

ptoffy: Isn't it more swifty to let it be inferred from the LHS? Explicitly if needed, e.g.
```swift
let result: [UInt8] = encode(...)
```

Package.swift
adam-fowler: Mutex?

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
adam-fowler: If you can infer it, that's great, but `result = encode(...)` does not provide enough detail. I personally think it is more swifty to infer what `result` is from the function call than infer the function that should be called based off what I expect it to return.

Package.swift
ptoffy: Well there's also using Swift Testing

Sources/MultipartKit/FormDataEncoder/FormDataEncoder.swift
ptoffy: Cool - I guess it's more of a style thing but I'm not precious about it so happy to take your suggestion

Sources/MultipartKit/FormDataEncoder/Storage.swift
ptoffy: Well `Storage` is shared across the various encoding containers and that might lead to issues

Sources/MultipartKit/FormDataDecoder/FormDataDecoder+Decoder.swift
ptoffy: We can probably remove `Sendable` annotations across the decoders but encoding shares the storage state so I doubt that will even compile with Swift 6

Sources/MultipartKit/FormDataEncoder/Storage.swift
ptoffy: I mean I'm open to suggestions if you believe this can be solved otherwise

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
ptoffy: No, I mean we can't pass the raw pointer to the decode method because it expects the collection of bytes, so this can't be done
```swift
try string.utf8.withContiguousStorageIfAvailable { bytes in
    decode(D.self, from: bytes, boundary: boundary)
} ?? decode(D.self, from: Array(string.utf8), boundary: boundary)
```
And if we were to do something like 
```swift
if let bytes = string.utf8.withContiguousStorageIfAvailable(Array.init) {
    try decode(D.self, from: bytes, boundary: boundary)
} else {
    try decode(D.self, from: Array(string.utf8), boundary: boundary)
}
```
we're still initialising an array with the raw bytes so this doesn't really save us a copy. 
Unless you mean a different way of using `withContiguousStorageIfAvailable `

Sources/MultipartKit/MultipartParser+parse.swift
ptoffy: But why? We're just doing that once at the beginning, this is the sync parse

Sources/MultipartKit/MultipartParserAsyncSequence.swift
adam-fowler: Yes just one part of the message eg I have a Multipart message with a zip file in there plus some metadata. I want to save the zip files to disk (using the least amount of memory possible ie streaming it) but parse the metadata with Codable so need the whole of it in memory.

Sources/MultipartKit/FormDataEncoder/Storage.swift
adam-fowler: Is Storage accessed from multiple tasks? Encoders are normally single threaded

adam-fowler: You would remove the Sendable conformance completely. It is not needed. The gain is you remove all the locks.

ptoffy: Right. I would've expected Swift 6 language mode to complain but apparently we're fine

Sources/MultipartKit/MultipartParserAsyncSequence.swift
ptoffy: Added https://github.com/vapor/multipart-kit/pull/100/commits/38bc3c1d94b90556cbdebfa21cbfb12522d09c4b#diff-0d53ff899089448c57c2915f2315bbf9d9e50f9a16caa3e2669249e84ff87d35R76

Sources/MultipartKit/FormDataEncoder/Storage.swift
```diff
@@ -1,33 +1,38 @@
 import Collections
+import Synchronization
```
0xTim, line 2: What's this? And where do we use it?

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,264 @@
...
-        self.onPartComplete = { }
-
-        self.boundary = Array("\r\n--\(boundary)".utf8)
-        self.boundaryLength = self.boundary.count
-        self.state = .preamble()
-    }
-
-    public func execute(_ string: String) throws {
-        try execute(ByteBuffer(string: string))
+        self.boundary = [45, 45] + ArraySlice(boundary.utf8)
```
0xTim, line 26: Add a comment as to this magic number or change it to a static named variable

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,264 @@
...
-            switch (crlf, byte) {
-            case (.cr, .cr):
-                crlf = .lf
-            case (.lf, .lf):
-                return nil
-            default:
-                throw Error.syntax
+    private mutating func parseBody(from buffer: ArraySlice<UInt8>) -> ReadResult {
+        // read until CRLF
+        switch buffer.getFirstRange(of: [13, 10] + boundary) {
```
0xTim, line 90: Again magic numbers

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,264 @@
...
         }
-    }
-}
-
-private extension UInt8 {
-    static let colon: UInt8 = 58
-    static let lf: UInt8 = 10
-    static let cr: UInt8 = 13
-    static let hyphen: UInt8 = 45
-    static let space: UInt8 = 9
```
0xTim, line 263: Let's bring something like this back to make it clear what everything is

Sources/MultipartKit/MultipartParser.swift
```diff
@@ -1,318 +1,264 @@
...
-        CTL         = <any US-ASCII control character (octets 0 - 31) and DEL (127)>
-        tspecials   = "(" | ")" | "<" | ">" | "@"
-                    | "," | ";" | ":" | "\" | DQUOTE
-                    | "/" | "[" | "]" | "?" | "="
-                    | "{" | "}" | SP | HT
-        DQUOTE      = <US-ASCII double-quote mark (34)>
-        SP          = <US-ASCII SP, space (32)>
-        HT          = <US-ASCII HT, horizontal-tab (9)>
-     */
-    private static let allowedHeaderFieldNameCharacterFlags: [Bool] = [
```
0xTim, line 280: Where do we check for this now?

0xTim (changes requested):
Some comments. This is starting to look good and we're getting close to be able to tag an alpha I think. 

Would be good to get the performance tests set up so we can measure changes going forward and code coverage to see which edge cases we don't currently test

Sources/MultipartKit/FormDataEncoder/Storage.swift
adam-fowler: That'll be from when Paul was using a Mutex protected state, which got removed so this can go

Sources/MultipartKit/MultipartParserAsyncSequence.swift
```diff
@@ -0,0 +1,107 @@
...
+            }
+        }
+
+        public mutating func nextCollatedPart() async throws -> MultipartSection<BackingSequence.Element>? {
+            var headerFields = HTTPFields()
+
+            while let part = try await next() {
+                switch part {
+                case .headerFields(let fields):
+                    headerFields.append(contentsOf: fields)
```
adam-fowler, line 88: Is there no chance you can read a header part after having read body parts?

adam-fowler (commented):
Looking good. Maybe we should have two async sequences though. One that does the collation automatically and one that requires you to use `nextCollatedPart` when you want a collated part.

I would rename the current AsyncSequence `StreamingMultipartParserAsyncSequence` and add the following 
```swift
public struct MultipartParserAsyncSequence<BackingSequence: AsyncSequence>: AsyncSequence
where BackingSequence.Element: MultipartPartBodyElement & RangeReplaceableCollection {
    let streamingSequence: StreamingMultipartParserAsyncSequence<BackingSequence>

    public struct AsyncIterator: AsyncIteratorProtocol {
        public mutating func next() async throws -> MultipartSection<BackingSequence.Element>? {
            try await self.streamingIterator.nextCollatedPart()
        }

        var streamingIterator: StreamingMultipartParserAsyncSequence<BackingSequence>.AsyncIterator
    }

    public func makeAsyncIterator() -> AsyncIterator {
        return .init(streamingIterator: self.streamingSequence.makeAsyncIterator())
    }
}
```

Sources/MultipartKit/MultipartParserAsyncSequence.swift
```diff
@@ -0,0 +1,107 @@
...
+    public init(boundary: String, buffer: BackingSequence) {
+        self.parser = .init(boundary: boundary)
+        self.buffer = buffer
+    }
+
+    public func makeAsyncIterator() -> Iterator {
+        Iterator(parser: parser, iterator: buffer.makeAsyncIterator())
+    }
+
+    public struct Iterator: AsyncIteratorProtocol {
```
adam-fowler, line 41: You should probably call this `AsyncIterator` instead of `Iterator` which is the associated type for a Sequence

Sources/MultipartKit/MultipartParser.swift
ptoffy: Here via `HTTPTypes` https://github.com/vapor/multipart-kit/blob/v5/Sources/MultipartKit/MultipartParser.swift#L171-L173

Sources/MultipartKit/MultipartParserAsyncSequence.swift
```diff
@@ -0,0 +1,107 @@
...
+
+            while let part = try await next() {
+                switch part {
+                case .headerFields(let fields):
+                    headerFields.append(contentsOf: fields)
+                case .bodyChunk(let chunk):
+                    self.currentCollatedBody.append(contentsOf: chunk)
+                    if !headerFields.isEmpty {
+                        let returningFields = headerFields
+                        headerFields = .init()
```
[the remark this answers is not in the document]
ptoffy: Well if there's different parts in a message there might other headers later and this empties them before possibly reading other ones

Sources/MultipartKit/MultipartParserAsyncSequence.swift
ptoffy: Yes there is if the message has different parts

ptoffy: FYI in this case `fields` will always have just one field because the original sequence returns single header fields and body chunks. I'm not sure if this is misleading and should be clarified somehow, either with docs or with a name change

ptoffy:
I agree with the proposed changes but I moved the `nextCollatedPart()` to simply be the `next()` method of the new sequence which makes more sense to me

Sources/MultipartKit/MultipartSection.swift
```diff
@@ -0,0 +1,7 @@
+import HTTPTypes
+
+public enum MultipartSection<Body: MultipartPartBodyElement>: Equatable, Sendable {
```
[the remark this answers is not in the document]
ptoffy: I hear you but I think not having this is more trouble than worth. Right now we can do
```swift
switch part {
case .headerFields(let fields): print(fields)
case .bodyChunk(let chunk): print(chunk)
case .boundary: ...
}
```
If we made it a structure we'd have either
```swift
switch part {
case .headerFields(): print(part.headerFields!) // or guard unwrap checking if we're in the right case, which is silly
case .bodyChunk(): print(part.bodyChunk!) 
case .boundary(): ...
}
```
or something like
```swift
part.handle(
  onHeaderFields: { fields in 
    print(fields)
  },
  onBodyChunk: { chunk in 
    print(chunk)
  },
  onBoundary: { end in 
    ...
  }
)
```
and none of these is particularly appealing IMHO. But I'm open to suggestions if you know of non-invasive way to hide the enum

0xTim (approved):
Nothing blocking on my end. Once other comments have been resolved we can look at integrating this into projects to see how it actually works

Sources/MultipartKit/MultipartSection.swift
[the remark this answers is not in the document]
0xTim: Lets keep it for now and see how we get on with a TODO to review

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -1,58 +1,67 @@
...
         for part in parts {
-            buffer.writeString("--")
-            buffer.writeString(boundary)
-            buffer.writeString("\r\n")
-            for (key, val) in part.headers {
-                buffer.writeString(key)
-                buffer.writeString(": ")
-                buffer.writeString(val)
-                buffer.writeString("\r\n")
+            buffer.append(contentsOf: Array("--\(boundary)".utf8) + crlf)
```
Joannis, line 57: Instead of copying the boundary into an array, write the leading tokens, boundary and CRLF separately. Saves copies and allocs, thus a lot of work.

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -1,58 +1,67 @@
...
+            for field in part.headerFields {
+                buffer.append(contentsOf: Array("\(field.description)".utf8) + crlf)
```
Joannis, line 59: Same here, you can easily prevent a copy

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -1,58 +1,67 @@
...
-            buffer.writeBuffer(&body)
-            buffer.writeString("\r\n")
+            buffer.append(contentsOf: crlf)
+            buffer.append(contentsOf: part.body)
+            buffer.append(contentsOf: crlf)
         }
-        buffer.writeString("--")
-        buffer.writeString(boundary)
-        buffer.writeString("--\r\n")
+        buffer.append(contentsOf: Array("--\(boundary)--".utf8) + crlf)
```
Joannis, line 65: Unnecessary copy

Sources/MultipartKit/FormDataDecoder/FormDataDecoder.swift
Joannis: I would really prefer not copying unnecessarily in a library like this

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -1,58 +1,67 @@
...
+    ///   - parts: One or more ``MultipartPart``s to serialize into a buffer.
+    ///   - buffer: Buffer to write to.
+    /// - Throws: Any errors that may occur during serialization.
+    /// - Note: `ByteBuffer` directly won't work because we have no dependency on NIO.
+    ///   You can use `ByteBufferView` via `ByteBuffer.readableBytesView`.
+    public func serialize<OutputBody: MultipartPartBodyElement>(
+        parts: [MultipartPart<some MultipartPartBodyElement>],
+        into buffer: inout OutputBody
+    ) throws where OutputBody: RangeReplaceableCollection {
+        let crlf = Array("\r\n".utf8)
```
Joannis, line 55: CRLF is copied into this a lot. I would keep it as `"\r\n".utf8` and append those values. Don't copy it into an array if possible

Joannis (changes requested):
My main (only) issue is that this library makes a lot of copies out of every byte carrier (String, Array etc)

Sources/MultipartKit/MultipartSerializer.swift
ptoffy: Actually I did add in https://github.com/vapor/multipart-kit/blob/v5/Sources/MultipartKit/Utilities.swift#L61 so using that now

Sources/MultipartKit/FormDataDecoder/FormDataDecoder+SingleValueContainer.swift
```diff
@@ -0,0 +1,63 @@
...
+            case is Float.Type:
+                String(bytes: part.body, encoding: .utf8).flatMap(Float.init) as? T
+            case is Double.Type:
+                String(bytes: part.body, encoding: .utf8).flatMap(Double.init) as? T
+            case is Bool.Type:
+                String(bytes: part.body, encoding: .utf8).flatMap(Bool.init) as? T
+            case is Data.Type:
+                Data(part.body) as? T
+            case is URL.Type:
+                String(bytes: part.body, encoding: .utf8).flatMap(URL.init(string:)) as? T
```
gwynne, line 42:
```suggestion
                String(bytes: part.body, encoding: .utf8) as? T
            case let IntType as any FixedWidthInteger.Type:
                String(decoding: part.body, as: UTF8.self).flatMap(IntType.init) as? T
            case is Float.Type:
                String(decoding: part.body, as: UTF8.self).flatMap(Float.init) as? T
            case is Double.Type:
                String(decoding: part.body, as: UTF8.self).flatMap(Double.init) as? T
            case is Bool.Type:
                String(decoding: part.body, as: UTF8.self).flatMap(Bool.init) as? T
            case is Data.Type:
                Data(part.body) as? T
            case is URL.Type:
                String(decoding: part.body, as: UTF8.self).flatMap(URL.init(string:)) as? T
```
Also, for the "basic" types (`String`, integers, `Float`/`Double`, `Bool`) you get a significant speedup by implementing the individual non-generic `decode()` methods for those types rather than doing a runtime type check in the generic method.

Sources/MultipartKit/FormDataDecoder/FormDataDecoder+KeyedContainer.swift
```diff
@@ -18,10 +18,11 @@ extension FormDataDecoder.KeyedContainer: KeyedDecodingContainerProtocol {
         data.keys.contains(key.stringValue)
     }
 
-    func getValue(forKey key: any CodingKey) throws -> MultipartFormData {
+    func getValue(forKey key: any CodingKey) throws -> MultipartFormData<Body> {
         guard let value = data[key.stringValue] else {
             throw DecodingError.keyNotFound(
-                key, .init(
+                key,
+                DecodingError.Context(
```
gwynne, line 25:
```suggestion
                .init(
```

Sources/MultipartKit/FormDataEncoder/FormDataEncoder+SingleValueContainer.swift
```diff
@@ -0,0 +1,34 @@
...
+#else
+    import Foundation
+#endif
+
+extension FormDataEncoder.Encoder: SingleValueEncodingContainer {
+    func encodeNil() throws {
+        // skip
+    }
+
+    func encode<T: Encodable>(_ value: T) throws {
```
gwynne, line 12: Again, you see a speedup by implementing the individual non-generic `encode()` methods rather than doing runtime type checks in the generic method.

Sources/MultipartKit/StreamingMultipartParserAsyncSequence.swift
```diff
@@ -0,0 +1,74 @@
...
+        var iterator: BackingSequence.AsyncIterator
+
+        public mutating func next() async throws -> MultipartSection<BackingSequence.Element>? {
+            while true {
+                switch parser.read() {
+                case .success(let optionalPart):
+                    switch optionalPart {
+                    case .none: continue
+                    case .some(let part): return part
+                    }
```
gwynne, line 50:
```suggestion
                case .success(.some(let part)):
                    return part
                case .success(_):
                    continue
```

gwynne (changes requested):
Various nits

Sources/MultipartKit/StreamingMultipartParserAsyncSequence.swift
ptoffy: Not sure I'm a fan of this

gwynne (approved):

Sources/MultipartKit/StreamingMultipartParserAsyncSequence.swift
dimitribouniol: If anything, I would explicitly call out the `.none` in the second case to make it explicit. (Also, reminder that `case .success(let part?)` is somehow valid syntax nobody remembers to use)

Sources/MultipartKit/MultipartSerializer.swift
```diff
@@ -1,58 +1,75 @@
...
+    ///   - parts: One or more ``MultipartPart``s to serialize into a buffer.
+    ///   - buffer: Buffer to write to.
+    /// - Throws: Any errors that may occur during serialization.
+    /// - Note: `ByteBuffer` directly won't work because we have no dependency on NIO.
+    ///   You can use `ByteBufferView` via `ByteBuffer.readableBytesView`.
+    public func serialize<OutputBody: MultipartPartBodyElement>(
+        parts: [MultipartPart<some MultipartPartBodyElement>],
+        into buffer: inout OutputBody
+    ) throws where OutputBody: RangeReplaceableCollection {
         for part in parts {
```
MahdiBM, line 55: I haven't kept up with the PR but it looks like there is opportunity for some `buffer.reserveCapacity(minimumWritableBytes:)`s.

0xTim: I think I'd this point we should merge, add benchmarks (including allocations) and then iterate forward

MahdiBM: sounds good, since that will be an optimization too.

ptoffy: That's not so easy, the buffer isn't necessarily a `ByteBuffer`

adam-fowler (commented):
> I agree with the proposed changes but I moved the `nextCollatedPart()` to simply be the `next()` method of the new sequence which makes more sense to me

While I can see the sense in this, I'm thinking of how this may be used. Given multipart files can come with sections of different structure, some you might want to stream and some you might want collated I think you should re-instate the `nextCollatedPart()`.

One other thing to consider. Nobody is ever going to want to stream headers, so collating them automatically makes sense.

I am trying this out with my swift package registry example just now. Will report back. To give you some context SwiftPM packages a file into a Multipart file consisting of headers for metadata, a metadata block, a headers for the package zip, and then the zip file of the package. I don't want to stream the first three sections but will want to stream the last.

adam-fowler:
Yeah by making me deal with streaming all the sections it is painful. Would be great if I could do 
```swift
let multipartStream = StreamingMultipartParserAsyncSequence(...)
let iterator = multipartStream.makeAsyncIterator()
while let part = try await iterator.next() {
    guard case .headerFields(let headers) else { continue }
    // is metadata
    if getParameter(headers[.contentDisposition].first, parameter: "name") == "metadata" {
        guard let metadata = iterator.nextCollatedBlock() else { throw Error() }
        parseMetadata(metadata)
    }
    // is archive
    if getParameter(headers[.contentDisposition].first, parameter: "name") == "archive" {
        while case .bodyChunk(let buffer) = try await iterator.next() {
            streamArchivePartToDisk(buffer)
        }
    }
}
```

0xTim:
I think that's a reasonable request for the API. If you're happy @adam-fowler I think we should merge as is and then add the improved API going forward?

ptoffy:
@0xTim I think adding the change in this PR makes sense. I'll add it in this afternoon and then we can merge

ptoffy:
@adam-fowler I updated the code to work with your example and I can confirm this works
```swift
let stream = makeParsingStream(for: message)
var iterator = StreamingMultipartParserAsyncSequence(boundary: boundary, buffer: stream).makeAsyncIterator()

var accumulatedBody: [UInt8] = []

while let part = try await iterator.next() {
    guard case .headerFields(let fields) = part else { continue }

    if fields.getParameter(.contentDisposition, "name") == "id" {
        guard let id = try await iterator.nextCollatedPart() else {
            throw MultipartMessageError.unexpectedEndOfFile
        }
        #expect(id == .bodyChunk(ArraySlice("123e4567-e89b-12d3-a456-426655440000".utf8)))
    } else {
        while case .bodyChunk(let chunk) = try await iterator.next() {
            accumulatedBody.append(contentsOf: chunk)
        }
    }
}

#expect(accumulatedBody == pngData)
```
though I haven't added the test to the branch because I'm a bit reluctant in making `fields.getParameter()` public. We can always do that later if needed. If you're happy with this I'll merge now

PR #105, opened 2025-01-15
Use shared benchmark CI

MahdiBM:


ptoffy (approved):

.github/workflows/benchmark.yml
```diff
@@ -1,6 +1,8 @@
 name: benchmark
 on:
   workflow_dispatch:
+  push:
+    branches: [main]
```
MahdiBM, line 5: added this. This helps with caching, as well as keeping a history of the benchmarks.

caches have branch-rules. meaning you can only retrieve a cache if it was saved by the current branch, OR the primary branch of the repo.
So the first run of a benchmark event in CI won't have a cache if we don't run benchmarks on push to main.

Though we could workaround this even without push to main. For example manually running the workflow on the main branch so it saves a base cache.

.github/workflows/benchmark.yml
```diff
@@ -12,172 +14,8 @@ on:
...
-      - name: Exit with correct status
-        run: |
-          EXIT_CODE='${{ steps.compare-benchmark.outputs.exit-status }}'
-          echo "Previous exit code was: $EXIT_CODE"
-          exit $EXIT_CODE
+  benchmark:
+    # If this is a pull-request related event, make sure this is a pr-approval event or someone has rerun the job.
+    if: (startsWith(github.event_name, 'pull_request') && (github.event.review.state == 'approved' || github.run_attempt > 1)) || true
+    uses: vapor/ci/.github/workflows/run-benchmark.yml@mmbm-benchmark
+    secrets: inherit
```
MahdiBM, line 21: keeping the `if` condition here ... should i move it to the reusable CI?

.github/workflows/benchmark.yml
```diff
@@ -12,172 +14,8 @@ on:
...
-      - name: Exit with correct status
-        run: |
-          EXIT_CODE='${{ steps.compare-benchmark.outputs.exit-status }}'
-          echo "Previous exit code was: $EXIT_CODE"
-          exit $EXIT_CODE
+  benchmark:
+    # If this is a pull-request related event, make sure this is a pr-approval event or someone has rerun the job.
+    if: (startsWith(github.event_name, 'pull_request') != true) || (github.event.review.state == 'approved' || github.run_attempt > 1)
+    uses: vapor/ci/.github/workflows/run-benchmark.yml@mmbm-benchmark
+    secrets: inherit
```
MahdiBM, line 21: keeping the if condition here ... should i move it to the reusable CI?

.github/workflows/benchmark.yml
```diff
@@ -12,172 +14,8 @@ on:
...
-
-      - name: Exit with correct status
-        run: |
-          EXIT_CODE='${{ steps.compare-benchmark.outputs.exit-status }}'
-          echo "Previous exit code was: $EXIT_CODE"
-          exit $EXIT_CODE
+  benchmark:
+    # If this is a pull-request related event, make sure this is a pr-approval event or someone has rerun the job.
+    if: (startsWith(github.event_name, 'pull_request') != true) || (github.event.review.state == 'approved' || github.run_attempt > 1)
+    uses: vapor/ci/.github/workflows/run-benchmark.yml@mmbm-benchmark
```
MahdiBM, line 20: need to change `@mmbm-benchmark` to `@main` before merge

.github/workflows/benchmark.yml
```diff
@@ -12,172 +14,8 @@ on:
...
-      - name: Exit with correct status
-        run: |
-          EXIT_CODE='${{ steps.compare-benchmark.outputs.exit-status }}'
-          echo "Previous exit code was: $EXIT_CODE"
-          exit $EXIT_CODE
+  benchmark:
+    # If this is a pull-request related event, make sure this is a pr-approval event or someone has rerun the job.
+    if: github.run_attempt > 1 || github.event.review.state == 'approved' || (startsWith(github.event_name, 'pull_request') != true)
+    uses: vapor/ci/.github/workflows/run-benchmark.yml@mmbm-benchmark
+    secrets: inherit
```
MahdiBM, line 21: keeping the if condition here ... should i move it to the reusable CI?

.github/workflows/benchmark.yml
```diff
@@ -12,172 +14,7 @@ on:
...
-          message: ${{ env.PR_COMMENT }}
-          comment-tag: benchmark-ci-comment
-
-      - name: Exit with correct status
-        run: |
-          EXIT_CODE='${{ steps.compare-benchmark.outputs.exit-status }}'
-          echo "Previous exit code was: $EXIT_CODE"
-          exit $EXIT_CODE
+  benchmark:
+    if: github.run_attempt > 1 || github.event.review.state == 'approved' || startsWith(github.event_name, 'pull_request') != true
```
MahdiBM, line 18: keeping the if condition here ... should i move it to the reusable CI?

0xTim: No this should probably be in the reusable CI

0xTim (approved):

.github/workflows/benchmark.yml
gwynne: Disagree - different repos may not want to use the same conditions.

MahdiBM: hmm hard decision. We generally don't add the if condition to the reusable CI so I guess we can keep at that.
I can see it coming back to bite us in either case.

gwynne (approved):

PR #106, opened 2025-01-16
Remove useless `Equatable` constraint

ptoffy:
We can get rid of the `Equatable` requirement in the multiparty body type typealias since it was really just used in tests

### DIFF
--- a/Benchmarks/Parser/AsyncSyncSequence.swift
+++ b/Benchmarks/Parser/AsyncSyncSequence.swift
@@ -31,7 +31,7 @@ final class AsyncSyncSequence<Base: Sequence>: AsyncSequence {
     }
 
     func makeAsyncIterator() -> Iterator {
-        defer { self.base = nil } // release the reference so no CoW is triggered
+        defer { self.base = nil }  // release the reference so no CoW is triggered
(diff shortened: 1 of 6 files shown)


Benchmarks/Parser/AsyncSyncSequence.swift
```diff
@@ -31,7 +31,7 @@ final class AsyncSyncSequence<Base: Sequence>: AsyncSequence {
     }
 
     func makeAsyncIterator() -> Iterator {
-        defer { self.base = nil } // release the reference so no CoW is triggered
+        defer { self.base = nil }  // release the reference so no CoW is triggered
```
gwynne, line 34:
```suggestion
        defer { self.base = nil } // release the reference so no CoW is triggered
```

ptoffy: This is swift-format's doing

gwynne: ... That's bizarre, wrong, stupid, pointless, and so forth. Oh well.

gwynne (approved):

0xTim (approved):

PR #107, opened 2025-02-02
Simplify some benchmarks

MahdiBM:


Benchmarks/Parser/Parser.swift
```diff
@@ -61,19 +50,18 @@ let benchmarks: @Sendable () -> Void = {
             ]
         )
     ) { benchmark in
-        defer { benchmarkIterated += 1 }
+        let bufferStream = chunkedMessage.async
 
+        benchmark.startMeasurement()
```
MahdiBM, line 55: Use `benchmark.startMeasurement()` to simplify benchmarks and not have to reserve buffers upfront.

gwynne (approved):

PR #112, opened 2025-02-04
Make MultipartParserError Public

### LINKED ISSUE #111: Make `MultipartParserError` public
### Please summarize your feature request

Packages throws `MultipartParserError` but this type has package scope

### Describe the functionality you're seeking in more detail

No point having different parser errors is the user can't differentiate between them

### Have you considered any alternatives?

_No response_

nakul-py:
### Summary
This PR makes the MultipartParserError publicly accessible within the MultipartKit package.

Fixes #111


ptoffy (changes requested):
Thanks for this, however this won't roll as it is. If we make the enum public as it is it will be set in stone and we won't be able to update it without breaking the API. Could you follow this approach instead? https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/JWTError.swift

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,37 @@
...
+    case invalidBody(String)
+}
+
+public struct MultipartParserErrorWrapper: MultipartParserError {
+    public let type: MultipartParserErrorType
+    
+    public init(type: MultipartParserErrorType) {
+        self.type = type
+    }
 }
```
MahdiBM, line 37: I don't think this is what @ptoffy meant by asking to follow https://github.com/vapor/jwt-kit/blob/main/Sources/JWTKit/JWTError.swift 😅

You simply need to move the current enum to be the [`Base` enum](https://github.com/vapor/jwt-kit/blob/03f5013f0b547ce43abe45e7e90711303a3e5495/Sources/JWTKit/JWTError.swift#L10C9-L10C49), then add static properties (`static func` / `static let`) like the other code.

This means we'll be able to add new static functions as freely as we want in the future without breaking public API while adding new enum cases to public enums breaks the public API because users will have "error you need to use the new enum case" in their `switch enumValue` statements.

So TL;DR, please try to follow the other enum's structure as closely as possible, with all the conformances and all.

nakul-py: Thanks for your fast review :) 
- https://github.com/vapor/multipart-kit/pull/112/commits/e7bcb2ae14e5206dc2cbdafd84d02c32cbe9929d

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,24 @@
...
-    case invalidHeader(reason: String)
-    case invalidBody(reason: String)
+public struct MultipartParserError: Swift.Error, Equatable {
+    public let reason: Reason
+
+    private init(reason: Reason) {
+        self.reason = reason
+    }
+
+    public enum Reason: Equatable {
```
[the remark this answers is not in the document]
MahdiBM: Also please move the enum to the top of the `struct MultipartParserError` like in [this example](https://github.com/vapor/jwt-kit/blob/03f5013f0b547ce43abe45e7e90711303a3e5495/Sources/JWTKit/JWTError.swift#L10), for consistency.

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,24 @@
...
+        case invalidBoundary
+        case invalidHeader(reason: String)
+        case invalidBody(reason: String)
+    }
+
+    public static let invalidBoundary = MultipartParserError(reason: .invalidBoundary)
+    
+    public static func invalidHeader(_ reason: String) -> MultipartParserError {
+        return MultipartParserError(reason: .invalidHeader(reason: reason))
```
MahdiBM, line 19: since we changed the initializer:
```suggestion
        return MultipartParserError(.invalidHeader(reason: reason))
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,24 @@
...
+    }
+    public static func invalidBody(_ reason: String) -> MultipartParserError {
+        return MultipartParserError(reason: .invalidBody(reason: reason))
```
MahdiBM, line 22: since we changed the initializer:
```suggestion
        return MultipartParserError(.invalidBody(reason: reason))
```

nakul-py: will we use base: instead of reason:

nakul-py: done

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+
+    enum Base: Equatable {
+        case invalidBoundary
+        case invalidHeader(reason: String)
+        case invalidBody(reason: String)
+    }
+
+    let base: Base
+
+    private init(base: Base) {
```
MahdiBM, line 13: nit pick: use `_` like the other code:
```suggestion
    private init(_ base: Base) {
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+        self.base = base
+    }
+
+    public static let invalidBoundary = MultipartParserError(base: .invalidBoundary)
+
+    public static func invalidHeader(reason: String) -> MultipartParserError {
+        return MultipartParserError(base: .invalidHeader(reason: reason))
```
MahdiBM, line 20:
```suggestion
        return MultipartParserError(.invalidHeader(reason: reason))
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+    }
+    public static func invalidBody(reason: String) -> MultipartParserError {
+        return MultipartParserError(base: .invalidBody(reason: reason))
```
MahdiBM, line 23:
```suggestion
        return MultipartParserError(.invalidHeader(reason: reason))
```

MahdiBM (changes requested):
Final changes hopefully, from my side at least.

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+        case invalidBody(reason: String)
+    }
+
+    let base: Base
+
+    private init(base: Base) {
+        self.base = base
+    }
+
+    public static let invalidBoundary = MultipartParserError(base: .invalidBoundary)
```
MahdiBM, line 17:
```suggestion
    public static let invalidBoundary = MultipartParserError(.invalidBoundary)
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
 /// Technical parsing error, such as malformed data or invalid characters.
 /// This is mainly used by ``MultipartParser``.
-package enum MultipartParserError: Swift.Error, Equatable {
-    case invalidBoundary
-    case invalidHeader(reason: String)
-    case invalidBody(reason: String)
+public struct MultipartParserError: Swift.Error, Equatable {
+
+    enum Base: Equatable {
```
MahdiBM, line 5: consistent spacing:
```suggestion
    enum Base: Equatable {
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+    private init(_ base: Base) {
+        self.base = base
+    }
+
+    public static let invalidBoundary = MultipartParserError(.invalidBoundary)
+
+    public static func invalidHeader(reason: String) -> MultipartParserError {
+        return MultipartParserError(.invalidHeader(reason: reason))
+    }
+    public static func invalidBody(reason: String) -> MultipartParserError {
```
MahdiBM, line 22: consistent spacing:
```suggestion
    }

    public static func invalidBody(reason: String) -> MultipartParserError {
```

MahdiBM (commented):
Looks good to me other than these 2 spaces. @ptoffy please take a look.

Sources/MultipartKit/MultipartParserError.swift
nakul-py: done

MahdiBM (approved):

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+    let base: Base
+
+    private init(_ base: Base) {
+        self.base = base
+    }
+
+    public static let invalidBoundary = MultipartParserError(.invalidBoundary)
+
+    public static func invalidHeader(reason: String) -> MultipartParserError {
+        return MultipartParserError(.invalidHeader(reason: reason))
```
ptoffy, line 19:
```suggestion
        .init(.invalidHeader(reason: reason))
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
...
+    }
+    
+    public static func invalidBody(reason: String) -> MultipartParserError {
+        return MultipartParserError(.invalidBody(reason: reason))
```
ptoffy, line 23:
```suggestion
        .init(.invalidBody(reason: reason))    
```

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,25 @@
 /// Technical parsing error, such as malformed data or invalid characters.
 /// This is mainly used by ``MultipartParser``.
-package enum MultipartParserError: Swift.Error, Equatable {
-    case invalidBoundary
-    case invalidHeader(reason: String)
-    case invalidBody(reason: String)
+public struct MultipartParserError: Swift.Error, Equatable {
```
ptoffy, line 3:
```suggestion
public struct MultipartParserError: Swift.Error, Equatable, Sendable {
```

ptoffy (changes requested):
A few nits but looking good

ptoffy:
Also @nakul-py this should go through a `swift-format` pass ideally

Sources/MultipartKit/MultipartParserError.swift
MahdiBM: Error already conforms to Sendable IIRC.

adam-fowler:
With the error constructed like this there is no way to extract the reason associated values.

adam-fowler:
There is still an issue here in that I cannot easily catch these errors without know the reason string eg

```swift
do {
    try multipartFormThings()
} catch let error as MultipartParserError where error == .invalidHeader("I need to know what string this is") {
    ...
}
```

@ptoffy Are the associated strings in the error necessary? If they are can we do something like this https://github.com/apple/swift-nio/blob/main/Sources/NIOFileSystem/FileSystemError.swift where the code and reason are separate.

MahdiBM:
To be honest I'd prefer to just have an enum with a new case "case DO_NOT_USE_OR_ELSE".
Then we can keep the pattern matching and all the enum goodies (other than exhaustive switches).

adam-fowler:
> To be honest I'd prefer to just have an enum with a new case `case __DO_NOT_USE_OR_ELSE__`. Then we can keep the pattern matching and all the enum goodies (other than exhaustive switches).

How does adding a __DO_NOT_USE_OR_ELSE__ case help?

Sources/MultipartKit/MultipartParserError.swift
```diff
@@ -1,7 +1,63 @@
...
+        public static let invalidBoundary = Self(.invalidBoundary)
+        public static let invalidHeader = Self(.invalidHeader)
+        public static let invalidBody = Self(.invalidBody)
+
+        public var description: String {
+            base.rawValue
+        }
+    }
+
+    private struct Backing: Equatable, Sendable {
```
adam-fowler, line 26: Not sure why you need the Backing type. What is this giving us over just adding the following to MultipartParserError
```swift
let errorType: ErrorType
let reason: String?
```

nakul-py: I use the Backing type because it provides a single place to modify error-related data without exposing too much detail in MultipartParserError.
And If we later want to add more fields (e.g., identifier, underlyingError, or HTTPStatusCode), we can do so without cluttering the main struct.

@adam-fowler will you in the favour of  remove Backing type.

MahdiBM:
> How does adding a `__DO_NOT_USE_OR_ELSE__` case help?

Well if there is such a case, it solves the exhaustive switch problem. Then we can just mark the enum as public.
I saw it somewhere else before too ... an Apple repo I think ... I think even Fabian approved that? can't remember more.

0xTim:
> Well if there is such a case, it solves the exhaustive switch problem. Then we can just mark the enum as public.
> I saw it somewhere else before too ... an Apple repo I think ... I think even Fabian approved that? can't remember more.

I'd vote not to use this - it's not a particularly nice solution and means that any new cases get blackholed into that catch all that just causes confusion because now people have to switch over the `DO_NOT_USE_OR_ELSE` case. We should stick with the approach we've taken in other repos until non-frozen enums land

MahdiBM:
It does have downsides, but the upside is it's much easier to maintain and you can use associated values with each case in a much nicer way. Also get the pattern matching stuff like in catch blocks.

I would go this route personally, but I don't think y'all like this (I mean you mentioned it too) so I'm just trying to explain the approach.
For this PR we can go whatever sensible route you'd want, like what we have in JWTKit, PostgresNIO etc...

ptoffy (approved):

PR #118, opened 2025-02-06
Try pull_request_target event in benchmarks

MahdiBM:


.github/workflows/benchmark.yml
```diff
@@ -5,9 +5,9 @@ on:
     branches: [main]
   pull_request_review:
     types: [submitted]
-  pull_request:
+  pull_request_target:
     branches: [main]
-    types: [synchronize]
+    types: [opened, reopened, synchronize, ready_for_review]
```
MahdiBM, line 10: Nothing will get run anyway, but the job would show up so one can manually re-run the job if they want to benchmark sooner

gwynne (approved):

PR #126, opened 2025-02-13
Add Android to CI

fpseverino:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,4 +15,5 @@ jobs:
       with_windows: true
       with_musl: true
       ios_scheme_name: multipart-kit
+      with_android: true
     secrets: inherit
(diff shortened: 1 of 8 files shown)


ptoffy (changes requested):
Let's remove the extra indentation

fpseverino:
> Let's remove the extra indentation

Ok, I also need to change the Swift format rule `"indentConditionalCompilationBlocks"` to `false` though

MahdiBM:
Hmm I probably would want to use the indentConditionalCompilationBlocks settings in the jwt-kit PR too 🤔 👍

ptoffy (approved):

ptoffy:
Yes @MahdiBM I'd say so. Then we can also update the centralised `.swift-format`

PR #128, opened 2025-02-14
Fix decoding message when using brackets in the CD's name value

### LINKED ISSUE #123: Encoded dictionaries can't necessarily be decoded
### Describe the issue

A dictionary with a key containing an open square bracket will generate an encoding that the decoder will fail to decode at all. If the string contains a close square bracket, it will be silently dropped

### Vapor version

4.108.0

### Operating system and version

Debian Linux 12 on WSL

### Swift version

6.0.2

### Steps to reproduce

```swift
struct HasADict: Content {
	var hints: [String: String]
}
let foo = HasADict(hints: ["f]o[o-": "bar"])
let serialized = try FormDataEncoder().encode(foo, boundary: "hello")
print("serialized: \(serialized)")
let deserialized = try FormDataDecoder().decode(HasADict.self, from: serialized, boundary: "hello")
print("deserialized: \(deserialized)")
```

### Outcome

```
serialized: --hello
Content-Disposition: form-data; name="hints[f]o[o-]"

bar
--hello--
 
Data corrupted at path 'hints.fo'. Decoding caught in recursion loop.
```

### Additional notes

If the string starts with a closing square bracket and doesn't contain an opening bracket, the closing square bracket will be dropped from the key on decoding.

ptoffy:
Fixes #123

### DIFF
--- a/Sources/MultipartKit/MultipartFormData.swift
+++ b/Sources/MultipartKit/MultipartFormData.swift
@@ -10,12 +10,10 @@ enum MultipartFormData<Body: MultipartPartBodyElement>: Sendable {
     case nestingDepthExceeded
 
     init(parts: [MultipartPart<Body>], nestingDepth: Int) {
-        self = parts.reduce(into: .empty) { result, part in
-            result.insert(
-                part,
-                at: part.name.map(makePath) ?? [],
-                remainingNestingDepth: nestingDepth
-            )
+        self = .empty
+        for part in parts {
+            let path = part.name.map(makePath) ?? []
+            insert(part, at: path, remainingNestingDepth: nestingDepth)
         }
     }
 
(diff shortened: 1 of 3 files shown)


Sources/MultipartKit/MultipartFormData.swift
```diff
@@ -46,8 +44,37 @@ enum MultipartFormData<Body: MultipartPartBodyElement>: Sendable {
...
+            }
+            i = string.index(after: j)
+        default:
+            result[writeIndex].append(string[i])
+            i = string.index(after: i)
+        }
+    }
+
+    return result
 }
```
MahdiBM, line 78: @gwynne can this be optimized? I think `result` should be initialized to a base of `string` so we don't need to copy the string chars into it? Not sure, haven't taken a close look.

Maybe there is a function in foundation or somewhere that can do this properly for us?

gwynne: There are a half-dozen ways to optimize this, but given the range of likely/common inputs I sincerely doubt you'd ever notice the difference in practice, even granted that you might occasionally encounter a pathological input. I'd be more concerned with the fact that you don't validate that `j` is a valid index (i.e. is before the `endIndex`) before using it to subscript the string - An input that ends with a `[` will crash. Might wanna fix that.

gwynne (changes requested):

Sources/MultipartKit/MultipartFormData.swift
0xTim: Can we use proper variable names as well? Not a big fan of `i` and `j`

PR #133, opened 2025-05-13
Split `MultipartPartConvertible` into two

### LINKED ISSUE #131: Split `MultipartPartConvertible` into `MultipartPartDecodable` and `MultipartPartEncodable`
Also I think the init() should probably throw an error instead of being optional. This would allow for more detailed errors.

```swift
public protocol MultipartPartDecodable {
    init(multipart: MultipartPart) throws
}

public protocol MultipartPartEncodable {
    var multipart: MultipartPart { get throws }
}
```

ptoffy:
This splits the `MultipartPartConvertible` protocol into two separate ones for encoding and decoding. Naming could be also reconsidered to a `MultipartPartCodable` to follow along with Swift standards.

This also makes the `init` throw instead of fail and the `multipart` property throw instead of return `nil` for better error diagnostics.

Closes #131

### DIFF
--- a/Sources/MultipartKit/FormDataDecoder/FormDataDecoder+SingleValueContainer.swift
+++ b/Sources/MultipartKit/FormDataDecoder/FormDataDecoder+SingleValueContainer.swift
@@ -94,7 +94,7 @@ extension FormDataDecoder.Decoder: SingleValueDecodingContainer {
         let decoded =
             switch T.self {
             case let multipartConvertible as any MultipartPartConvertible.Type:
-                multipartConvertible.init(multipart: part) as? T
+                try multipartConvertible.init(multipart: part) as? T
             case is MultipartPart<Body>.Type:
                 part as? T
             case is Data.Type:
--- a/Sources/MultipartKit/FormDataEncoder/FormDataEncoder+SingleValueContainer.swift
+++ b/Sources/MultipartKit/FormDataEncoder/FormDataEncoder+SingleValueContainer.swift
@@ -47,7 +47,10 @@ extension FormDataEncoder.Encoder: SingleValueEncodingContainer {
     func encode<T: Encodable>(_ value: T) throws {
         switch value {
         case let multipartConvertible as any MultipartPartConvertible<Body>:
-            guard let multipart = multipartConvertible.multipart else {
+            let multipart: MultipartPart<Body>
+            do {
+                multipart = try multipartConvertible.multipart
+            } catch {
                 return try value.encode(to: self)
             }
             storage.dataContainer = SingleValueDataContainer(part: multipart)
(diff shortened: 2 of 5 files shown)


Sources/MultipartKit/MultipartPartConvertible.swift
```diff
@@ -1,17 +1,23 @@
...
-    var multipart: MultipartPart<Body>? { get }
-    init?(multipart: MultipartPart<some MultipartPartBodyElement>)
+    var multipart: MultipartPart<Body> { get throws }
+}
+
+public protocol MultipartPartDecodable {
+    init(multipart: MultipartPart<some MultipartPartBodyElement>) throws
 }
 
+public protocol MultipartPartConvertible<Body>: MultipartPartEncodable, MultipartPartDecodable where Body == Self.Body {}
```
adam-fowler, line 13:
```suggestion
public typealias MultipartPartConvertible<Body> = MultipartPartEncodable<Body> & MultipartPartDecodable<Body> 
```

Sources/MultipartKit/FormDataDecoder/FormDataDecoder+SingleValueContainer.swift
```diff
@@ -94,7 +94,7 @@ extension FormDataDecoder.Decoder: SingleValueDecodingContainer {
         let decoded =
             switch T.self {
             case let multipartConvertible as any MultipartPartConvertible.Type:
```
adam-fowler, line 96:
```suggestion
            case let multipartDecodable as any MultipartPartDecodable.Type:
```

Sources/MultipartKit/FormDataEncoder/FormDataEncoder+SingleValueContainer.swift
```diff
@@ -47,7 +47,10 @@ extension FormDataEncoder.Encoder: SingleValueEncodingContainer {
     func encode<T: Encodable>(_ value: T) throws {
         switch value {
         case let multipartConvertible as any MultipartPartConvertible<Body>:
```
adam-fowler, line 49:
```suggestion
        case let multipartEncodable as any MultipartPartEncodable<Body>:
```

adam-fowler (commented):
If you are going to split convertible into decodable and encodable you might as well use that in the decoder and encoder, otherwise you are still forcing people to use convertible all the time.

Sources/MultipartKit/MultipartPartConvertible.swift
ptoffy: I've tried this but it doesn't work unfortunately because of `MultipartPart` conformance: if I leave the conformance as is it requires the `Body` generic argument and if I add them I get
> Cannot inherit from protocol type with generic argument 'MultipartPartConvertible<Body>'
however the protocol is a good solution in my opinion

PR #136, opened 2025-05-16
Replace `MultipartSerializer` with `MultipartWriter`

ptoffy:
This adds a `MultipartWriter` which replaces `MultipartSerializer`. The new structure follows the pattern adopted by Hummingbird's (and therefore Vapor 5's) `ResponseBody` structure.
This allows users to be able to define an arbitrary writing option (e.g write to a buffer, file IO, network connection etc).
The PR also introduces a default generic and synchronous `BufferedMultipartWriter` which writes the bytes to an in-memory buffer for the simplest situations. 
The `FormDataEncoder` also adopts the new structure, allowing for `MultipartSerializer` to be removed


Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,129 @@
...
+            buffer.append(contentsOf: field.name.rawName.utf8)
+            buffer.append(.colon)
+            buffer.append(.space)
+            buffer.append(contentsOf: field.value.utf8)
+            buffer.append(contentsOf: ArraySlice.crlf)
+        }
+        buffer.append(contentsOf: ArraySlice.crlf)
+    }
+
+    public mutating func writeBodyChunk(_ chunk: some MultipartPartBodyElement) async throws {
```
adam-fowler, line 70: `@inlinable`
Although is this function necessary if it is already defined in the extension to MultipartWriter. It doesn't provide any optimisation. If the correct functions are tagged as `@inlinable` the compiler will optimise the function down to the same thing.

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,129 @@
...
+    public mutating func writeBoundary(end: Bool = false) async throws {
+        buffer.reserveCapacity(boundary.utf8.count + 10)
+        buffer.append(.hyphen)
+        buffer.append(.hyphen)
+        buffer.append(contentsOf: boundary.utf8)
+        if end {
+            buffer.append(.hyphen)
+            buffer.append(.hyphen)
+        }
+        buffer.append(contentsOf: ArraySlice.crlf)
```
adam-fowler, line 55: Why don't you cache the full boundary instead of recreating it every time

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,129 @@
...
+            buffer.append(contentsOf: ArraySlice.crlf)
+        }
+        buffer.append(contentsOf: ArraySlice.crlf)
+    }
+
+    public mutating func writeBodyChunk(_ chunk: some MultipartPartBodyElement) async throws {
+        buffer.append(contentsOf: chunk)
+    }
+
+    public mutating func writeBodyChunks(_ chunks: some Sequence<some MultipartPartBodyElement>) async throws {
```
adam-fowler, line 74: See above above whether this is needed

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+    ///
+    /// This method should be called when all parts have been written to properly
+    /// terminate the multipart message.
+    ///
+    /// - Throws: Any error that occurs during writing.
+    mutating func finish() async throws
+}
+
+extension MultipartWriter {
+    private var boundaryPrefix: OutboundBody {
```
adam-fowler, line 68: This and `boundarySuffix` are not used anywhere

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+        var suffix = OutboundBody()
+        suffix.append(contentsOf: ArraySlice.twoHyphens)
+        return suffix
+    }
+
+    /// Writes a multipart boundary with optional termination.
+    ///
+    /// - Parameter end: Whether this is the final boundary that terminates the multipart message.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writeBoundary(end: Bool = false) async throws {
```
adam-fowler, line 85: `@inlinable`

Plus would it be sensible to cache the full boundary. Or at least provide a function that generates it so that types that conform to `MultipartWriter` can cache it and use it instead of writing each individual part

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+        }
+        boundaryBytes.append(contentsOf: ArraySlice<UInt8>.crlf)
+        try await write(bytes: boundaryBytes)
+    }
+
+    /// Writes HTTP header fields for a multipart part.
+    ///
+    /// - Parameter httpFields: The header fields to write.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writeHeaders(_ httpFields: HTTPFields) async throws {
```
adam-fowler, line 101: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+        }
+        bytes.append(contentsOf: ArraySlice.crlf)
+        try await write(bytes: bytes)
+    }
+
+    /// Writes a single body chunk.
+    ///
+    /// - Parameter chunk: The body chunk to write.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writeBodyChunk(_ chunk: some MultipartPartBodyElement) async throws {
```
adam-fowler, line 124: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+        try await write(bytes: chunk)
+    }
+
+    /// Writes multiple body chunks followed by a CRLF sequence.
+    ///
+    /// - Parameter chunks: A sequence of body chunks to write.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writeBodyChunks(_ chunks: some Sequence<some MultipartPartBodyElement>) async throws {
```
adam-fowler, line 132: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+            try await write(bytes: chunk)
+        }
+        try await write(bytes: ArraySlice.crlf)
+    }
+
+    /// Writes body chunks from an async sequence followed by a CRLF sequence.
+    ///
+    /// - Parameter chunks: An async sequence of body chunks to write.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writeBodyChunks<Chunks: AsyncSequence>(_ chunks: Chunks) async throws
```
adam-fowler, line 143: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+            try await write(bytes: chunk)
+        }
+        try await write(bytes: ArraySlice.crlf)
+    }
+
+    /// Writes a complete multipart part including boundary, headers, and body.
+    ///
+    /// - Parameter part: The multipart part to write.
+    /// - Throws: Any error that occurs during writing.
+    public mutating func writePart(_ part: MultipartPart<some MultipartPartBodyElement>) async throws {
```
adam-fowler, line 155: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
```diff
@@ -0,0 +1,181 @@
...
+        try await write(bytes: serializedPart)
+    }
+
+    /// Writes the final boundary to the multipart data.
+    ///
+    /// This method should be called when all parts have been written to properly
+    /// terminate the multipart message.
+    ///
+    /// - Throws: Any error that occurs during writing.
+    public mutating func finish() async throws {
```
adam-fowler, line 178: `@inlinable`

Sources/MultipartKit/Writer/MultipartWriter.swift
adam-fowler: If you are going to override this in types that conform to this type you should really add it as a protocol requirement. Similarly the same should be done with `writeBoundary`

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,129 @@
...
+        buffer.reserveCapacity(chunks.underestimatedCount * 64 + ArraySlice.crlf.count)
+        for chunk in chunks {
+            buffer.append(contentsOf: chunk)
+        }
+        buffer.append(contentsOf: ArraySlice.crlf)
+    }
+
+    public mutating func writePart(_ part: MultipartPart<some MultipartPartBodyElement>) async throws {
```
adam-fowler, line 82: See above above whether this is needed

ptoffy:
@adam-fowler I've applied most suggestions, thanks for the review!

I removed the custom implementations in the buffered writer and got somewhere near the same performance using the default ones in the protocol extension. 
I don't think we should add any methods (such as `writeHeaders` etc) to the protocol because of both a protocol responsibility principle and to not force any unnecessary low(er)-level implementation to the user. I'm aware that with the extension users wouldn't _have_ to implement the methods, but even if the method is not in the protocol, they can still override the default one with the correct signature. I think that's a good compromise for somebody looking to optimise performance, and it keeps the protocol simple.
In fact, in the `BufferedMultipartWriter`, `writePart` has been overridden with the slightly more efficient non-async version that we need anyway to implement `FormDataEncoder` synchronously.

As for caching the boundary bytes, at least in terms of allocations, there's no real optimisation with caching, I've tried it out on both `BufferedMultipartWriter` and the `EmbeddedWriter` in the `StreamingMultipartWriterAsyncSequence`. In fact there's a couple more allocations due to the fact that we have to store the boundary.  Implementors of `MultipartWriter` can still cache it themselves and override `writeBoundary` if needed. Because of this, I wouldn't be against adding a utility to make it easy to create the boundary if needed, something like 
```swift
extension MultipartWriter {
    public static func makeBoundaryBytes(_ boundary: String, end: Bool = false) -> OutboundBody {
        var bytes = OutboundBody()
        bytes.append(contentsOf: ArraySlice.twoHyphens)
        bytes.append(contentsOf: boundary.utf8)
        if end {
            bytes.append(contentsOf: ArraySlice.twoHyphens)
        }
        bytes.append(contentsOf: ArraySlice.crlf)
        return bytes
    }
}
```
what do you think?

adam-fowler:
> but even if the method is not in the protocol, they can still override the default one with the correct signature.

You might want to check that again. Remember my this/that example. When I had a generic `This` it chose the implementation in the `This` extension and not the override in the `That` type.

> Actually thinking more about it, not being able to use instance methods when using generics and existential might be an issue. Having the methods in the protocol might be the smarter thing to do

Saw this after writing the above

> In fact, in the `BufferedMultipartWriter`, `writePart` has been overridden with the slightly more efficient non-async version that we need anyway to implement `FormDataEncoder` synchronously.

Now I understand why you were so keen to keep the non-async implementation

> Because of this, I wouldn't be against adding a utility to make it easy to create the boundary if needed, something like

Looks good

adam-fowler (commented):
I'm happy for this as is, but because the default implementations of writeHeader/Boundary has loads of tiny writes I think a helper writer type which buffers the writes into a fixed size memory buffer, flushing to an underlying writer whenever it overflows would be helpful.

Unfortunately the best name I can think for this type would be `BufferedMultipartWriter`. Maybe the orginal `BufferedMultipartWriter`could be named `MultipartMemoryBufferWriter`.

ptoffy:
@adam-fowler added new `BufferedMultipartWriter` with your requested behaviour and renamed the old one to `MemoryMultipartWriter`

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,109 @@
...
+
+    /// Creates a new buffered multipart writer with a user-supplied buffer.
+    ///
+    /// - Parameters:
+    ///   - boundary: The boundary string to use for separating multipart parts.
+    ///   - buffer: The buffer to write to before flushing to the underlying writer.
+    ///   - bufferCapacity: Maximum number of bytes to buffer before writing to the underlying writer.
+    ///   - underlyingWriter: The writer that will receive the buffered data when capacity is reached.
+    @inlinable
+    public init(boundary: String, buffer: OutboundBody, bufferCapacity: Int, underlyingWriter: UnderlyingWriter) {
```
adam-fowler, line 68: Isn't bufferCapacity defined by buffer.count

Tests/MultipartKitTests/WriterTests.swift
```diff
@@ -0,0 +1,450 @@
...
+        // Write enough to exceed bufferCapacity, should flush
+        try await writer.write(bytes: ArraySlice("67890".utf8))
+        #expect(!countingWriter.writes.isEmpty)
+
+        // Write more, should buffer again
+        let writesAfterFlush = countingWriter.writes.count
+        try await writer.write(bytes: ArraySlice("abc".utf8))
+        #expect(countingWriter.writes.count == writesAfterFlush)
+
+        try await writer.finish()
```
adam-fowler, line 356:
```suggestion
        try await writer.finish(writingEndBoundary: false)
```

Tests/MultipartKitTests/WriterTests.swift
```diff
@@ -0,0 +1,450 @@
...
+        #expect(mockWriter.writeCallCount == 0)
+
+        let largeData = ArraySlice(Array(repeating: UInt8(65), count: 150))
+        try await writer.write(bytes: largeData)
+        let countAfterFirstWrite = mockWriter.writeCallCount
+
+        #expect(mockWriter.writeCallCount > 0)
+        #expect(mockWriter.lastWrittenData != nil)
+
+        try await writer.finish()
```
adam-fowler, line 253:
```suggestion
        try await writer.finish(writingEndBoundary: false)
```

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
ptoffy: I'd expect count and capacity to be two different things: the count is the actual number of elements and capacity is the max. Usually you'd pass in an empty buffer so the count will be 0

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,109 @@
...
+    ///
+    /// This method accumulates the provided bytes in the internal buffer. If the buffer's
+    /// capacity is reached, the entire buffer is forwarded to the underlying writer and then cleared.
+    ///
+    /// - Parameter bytes: The bytes to write to the buffer.
+    @inlinable
+    public mutating func write(bytes: some Collection<UInt8> & Sendable) async throws {
+        if currentBufferCount + bytes.count >= bufferCapacity {
+            try await underlyingWriter.write(bytes: self.buffer)
+            try await underlyingWriter.write(bytes: bytes)
```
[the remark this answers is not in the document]
ptoffy: In your example there's no real buffering no? You're just writing the buffer to the underlying writer every time and conditionally writing the new bytes directly to the underlying writer or just waiting for the next `write` to be called. I though the point of this was to decrease the writes to the underlying writer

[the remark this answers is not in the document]
ptoffy: Ah cool yeah this makes more sense

0xTim (approved):
Jumping in late so potentially missing some context - no major objections from me and looks good, just a query around the writer API

Sources/MultipartKit/Writer/BufferedMultipartWriter.swift
```diff
@@ -0,0 +1,118 @@
...
+            currentBufferCount += bytes.count
+        }
+    }
+
+    /// If the buffer has not been emptied by the last write,
+    /// flushes the final part of the message to the underlying writer.
+    ///
+    /// By default, writes the end boundary as required by the multipart protocol.
+    @inlinable
+    public mutating func finish(writingEndBoundary: Bool = true) async throws {
```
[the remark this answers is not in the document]
adam-fowler: Yeah a `withWriter` function might work. Depends on whether you want to pass the writer about. Can't think of any instance where that'd be needed. You could at least provide this as a high level API while leaving the others available.

PR #142, opened 2025-09-23
MultipartPartDecodable initialized with associatedtype Body

adam-fowler:
Use associatedtype `Body` in initialiser for `MultipartPartDecodable`. This also required setting this as the primary associated type so we can cast to it in `FormDataDecoder`.  This required tests to also be updated so the same buffer type was used in the `File` type and the test `decodeSimilVaporFileType` that used it.

Use associatedtype `Body` in initialiser for `MultipartPartDecodable`


ptoffy (commented):
If I remember correctly I originally went with `some MultipartPartBodyElement` instead of `Body` in the `init` to allow different types to be used in the encoder and decoder when conforming to `MultipartPartConvertible`. This breaks that. The associated type was probably left as an oversight

ptoffy (approved):

PR #145, opened 2025-11-18
Add streamable multipart part

ptoffy:
This adds a `StreamingMultipartPart` which is a `MultipartPart` but with a streamable body. In addition, there's also two new sequences, one for producing `StreamingMultipartPart`s from `MultipartSection`s and another one for the opposite direction. This PR also makes most sequences `Sendable` so they're easier to be used in modern concurrency codebases


Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
```diff
@@ -0,0 +1,130 @@
...
+        pendingBodyChunk: BodyChunk? = nil
+    ) {
+        self.backingIterator = makeBackingIterator()
+        self.pendingBodyChunk = pendingBodyChunk
+        self.stateMachine = .init()
+        self.isReading = false
+    }
+
+    func nextPart() async throws -> Element {
+        precondition(!isReading, "Streaming multipart message was iterated concurrently")
```
0xTim, line 27: Should we throw here, _just_ in case?

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
```diff
@@ -0,0 +1,130 @@
...
+                    return StreamingMultipartPart(headerFields: headerFields, body: body)
+                }
+            }
+        }
+
+        return nil
+    }
+
+    func nextBodyChunkForSubsequence(id: Int) async throws -> BodyChunk? {
+        precondition(!isReading, "StreamingMultipartPart body was iterated concurrently")
```
0xTim, line 90: As above

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
```diff
@@ -0,0 +1,130 @@
...
+            next = try await iterator.next(isolation: self)
+        } else {
+            next = try await iterator.next()
+        }
+
+        guard let next else { return nil }
+
+        switch next {
+        case .headerFields:
+            return nil
```
0xTim: This would be malformed input here right? If we encounter errors during body parsing? Do we want to surface an error here instead?

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
ptoffy: Hmm this is a programmer error, I'd lean more towards keeping it a precondition

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
ptoffy: That's a great spot. It's technically not reachable unless a user feeds the sequence hand-built parts with sections in the wrong order, but we should still guard against it and close the state machine

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
0xTim: Basically is there anyway at all a file can be manipulated to hit this path? I don't think so, in which case I'm ok keeping it

0xTim (approved):
One thing for clarification, otherwise LGTM

Sources/MultipartKit/StreamingMultipartPart/StreamingMultipartPart+SharedIterator.swift
ptoffy: No, it doesn't depend on user input

PR #150, opened 2026-07-09
Update to Swift 6.1

ptoffy:
This updates to Swift 6.1, adopts the FoundationEssentials and adds some Swift settings to the package

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -1,4 +1,4 @@
-// swift-tools-version:6.0
+// swift-tools-version:6.1
 import PackageDescription
 
 let package = Package(
@@ -41,6 +41,10 @@ let package = Package(
 
 var swiftSettings: [SwiftSetting] {
     [
-        .enableUpcomingFeature("ExistentialAny")
+        .enableUpcomingFeature("ExistentialAny"),
+        //.enableUpcomingFeature("InternalImportsByDefault"),
+        .enableUpcomingFeature("MemberImportVisibility"),
+        .enableUpcomingFeature("InferIsolatedConformances"),
+        // .enableUpcomingFeature("NonisolatedNonsendingByDefault"),
     ]
 }
--- a/Package.swift
+++ b/Package.swift
@@ -41,10 +41,14 @@ let package = Package(
 
 var swiftSettings: [SwiftSetting] {
     [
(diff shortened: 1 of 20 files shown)


Package.swift
```diff
@@ -41,6 +41,14 @@ let package = Package(
 
 var swiftSettings: [SwiftSetting] {
     [
-        .enableUpcomingFeature("ExistentialAny")
+        .enableExperimentalFeature("SuppressedAssociatedTypesWithDefaults"),
+        .enableExperimentalFeature("LifetimeDependence"),
+        .enableExperimentalFeature("Lifetimes"),
+        .enableUpcomingFeature("LifetimeDependence"),
```
MahdiBM, line 47: We're not leveraging any of these?

Sources/MultipartKit/Parser/StreamingMultipartParserAsyncSequence.swift
```diff
@@ -79,10 +79,17 @@ where BackingSequence.Element: MultipartPartBodyElement {
...
                     let next: BackingSequence.Element?
                     do {
-                        next = try await iterator.next()
+                        if #available(macOS 15.0, iOS 18.0, watchOS 11.0, tvOS 18.0, visionOS 2.0, *) {
+                            next = try await backingIterator.next(isolation: #isolation)
+                        } else {
+                            nonisolated(unsafe) var iterator = backingIterator
+                            defer { backingIterator = iterator }
+                            next = try await iterator.next()
+                        }
```
MahdiBM, line 88: These `nonisolated(unsafe)`s are actually safe I'd assume?

MahdiBM (approved):
Looks good; some minor issues:

Sources/MultipartKit/Parser/StreamingMultipartParserAsyncSequence.swift
ptoffy: Yeah had to add it because of `NonisolatedNonsendingByDefault`. I think the only situation this would be unsafe would be if two actors consumed the iterator concurrently but that's not valid

Package.swift
ptoffy: Yeah this was @0xTim's advice to keep all of them for consistency and future-proofing

PR #152, opened 2026-07-10
Add some performance improvements

ptoffy:
This adds some performance improvements:
- in `MultipartParser`: don't recompute `CRLF` + `boundary` for each parse since the boundary is constant for the parser
- in `MultipartFormData`: avoid keeping two references of the `array` and `dictionary` which trigger CoW
- in `MemoryMultipartWriter`: call `reserveCapacity` _before_, from outside the writer, to allocate only once


MahdiBM (approved):
nice

PR #154, opened 2026-07-13
Introduce `MultipartWriter.Failure`

### LINKED ISSUE #138: Introduce `MultipartWriterError`
### Please summarize your feature request

`MultipartWriter` & friends should throw `MultipartWriterError`s only

### Describe the functionality you're seeking in more detail

Just like we have `MultipartParserError` in the parsing methods, the writer structures should have an appropriate `MultipartWriterError` too.

### Have you considered any alternatives?

_No response_

ptoffy:
This adds a `Failure` `associatedtype` to the `MultipartWriter` in the style of https://github.com/swiftlang/swift-evolution/blob/main/proposals/0421-generalize-async-sequence.md, allowing for non-throwing writers to be called without  an insincere `try`. This is better than introducing a separate `MultipartWriterError`, as our writer protocol doesn't really ever throw (in contrast to the parser) so we can leave it up to the conforming writer to decide whether they need to. 

We cannot use this in the AsyncSequences quite yet since we'd need macOS 15+ to reference `AsyncSequence.Failure`


0xTim (approved):

ptoffy:
Merging over the 6.1 Windows failure (https://github.com/swiftlang/swift/issues/90657#issuecomment-4970911211) and the API break which is expected

